Issue #6, opened 2022-05-26
Setup Django back-end


caseycesari:
Consult other recent projects and add a Django and DRF back-end. 

Add a demo endpoint to the DRF API that demonstrates that the back-end is working.

Issue #7, opened 2022-05-26
Fix cibuild script


caseycesari:
This script is partly broken because the project does not have a Django back-end yet. This will be added in #6.

However, there appears to be a problem with the front-end setup as well. Here's the error I get when I run `cibuild` locally.

```
 ./scripts/cibuild
Creating iow-boundary-tool_shellcheck_run ... done
Creating iow-boundary-tool_app_run ... done
Creating iow-boundary-tool_app_run ... done
yarn run v1.22.18
$ craco test --watchAll=false
 FAIL  src/App.test.js
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading '_cracoConfig')

      at Object.process (node_modules/@craco/craco/lib/features/jest/jest-babel-transform.js:12:74)
      at ScriptTransformer.transformSource (node_modules/@jest/transform/build/ScriptTransformer.js:619:31)
      at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:765:40)
      at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:822:19)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        1.352 s
Ran all test suites.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
ERROR: 1
```

Github Actions reports a different error with the front-end: https://github.com/azavea/iow-boundary-tool/runs/6598656396?check_suite_focus=true

Fix this error so that the cibuild script can successfully execute locally and on CI.

PR #9, opened 2022-06-08
Create Django Backend

nanotubing:
## Overview

This work has been to create the django backend and creates an example endpoint to demonstrate the functionality

Connects #6 

### Demo

```
$ http http://localhost:8181/api/example/

HTTP/1.1 200 OK
Connection: close
Content-Length: 20
Content-Type: text/plain; charset=UTF-8
Date: Wed, 08 Jun 2022 21:03:14 GMT
Referrer-Policy: same-origin
Server: gunicorn
X-Content-Type-Options: nosniff
X-Frame-Options: DENY

Example API response
```

## Testing Instructions

 * check out the branch
 * run `./scripts/update`
 * run `./scripts/server`
 * test the API endpoint as seen above using httpie, or some other API testing utility


src/django/iow/settings.py
```diff
@@ -0,0 +1,237 @@
...
+
+INSTALLED_APPS = [
+    "whitenoise.runserver_nostatic",
+    'django.contrib.admin',
+    'django.contrib.auth',
+    'django.contrib.contenttypes',
+    "django.contrib.gis",
+    'django.contrib.sessions',
+    'django.contrib.messages',
+    # 'django.contrib.staticfiles',
```
caseycesari, line 94: Is this commented out because of the use of Django-SPA?

docker-compose.override.yml
```diff
@@ -3,3 +3,6 @@ services:
   app:
     ports:
       - 4545:4545
+  django:
+    ports:
+      - 8181:8181
```
caseycesari, line 8: Can you add this port to the README port section?

caseycesari (approved):
Looking good! Just a couple questions.

caseycesari:
Also FYI #7 is to address the CI failure.

src/django/iow/settings.py
nanotubing: I tried setting up the static files stuff, but was getting errors, so I figured I'd leave it commented out for now, and circle back to it later since we're not trying to serve any static content right now. I'll open an issue to fix this

caseycesari: Sounds good. Thanks!

src/django/iow/urls.py
```diff
@@ -0,0 +1,24 @@
...
+    1. Import the include() function: from django.urls import include, path
+    2. Add a URL to urlpatterns:  path('blog/', include('blog.urls'))
+"""
+from django.contrib import admin
+from django.urls import path, include
+
+urlpatterns = [
+    path('admin/', admin.site.urls),
+    path('api/', include('api.urls')),
+    path('health-check/', include('watchman.urls')),
```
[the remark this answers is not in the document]
nanotubing: will do

Issue #10, opened 2022-06-10
Fix static content hosting


nanotubing:
as part of #6, the static content hosing provided by whitenoise and staticfiles was not fully configured, and some code remains commented out.
1. uncomment or remove all commented out code
2. configure static content hosting in advance of frontend creation

PR #14, opened 2022-07-27
Enable Static File hosting

nanotubing:
Cleaning up configuration to enable django hosting static files from the frontend.

## Overview

This PR configures the plugins and configuration necessary for django to host static files for the frontend.

Connects #10 

## Testing Instructions

 * Check out this branch
 * Run `scripts/update`
 * Run `scripts/server`
 * In a web browser, load the [local development site](http://localhost:8181/)
 * If this test is successful, it will load a blank white screen with the title 'React App'. If it is not successful, it will load a "Page not found" 404 error

### DIFF
--- a/scripts/lint
+++ b/scripts/lint
@@ -23,7 +23,6 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
             scripts/*
 
         # Lint JavaScript
-        # TODO turn this back on when there's javascript to check
         docker-compose \
             run --rm --entrypoint ./node_modules/.bin/eslint \
             app src/ --ext .js --ext .jsx
--- a/scripts/lint
+++ b/scripts/lint
@@ -24,9 +24,9 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
 
         # Lint JavaScript
         # TODO turn this back on when there's javascript to check
-        # docker-compose \
-        #     run --rm --entrypoint ./node_modules/.bin/eslint \
-        #     app src/ --ext .js --ext .jsx
+        docker-compose \
+            run --rm --entrypoint ./node_modules/.bin/eslint \
+            app src/ --ext .js --ext .jsx
 
         if [[ -n "${CI}" ]]; then
             # Lint Django app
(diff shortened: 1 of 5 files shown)


scripts/lint
```diff
@@ -24,9 +24,9 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
 
         # Lint JavaScript
         # TODO turn this back on when there's javascript to check
```
caseycesari, line 26: Stale comment

scripts/update
```diff
@@ -38,6 +38,6 @@ if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
 
         # Collect Django static files
         # TODO enable this after static file hosting is corrected
```
caseycesari, line 40: Stale comment

src/django/iow/settings.py
```diff
@@ -216,9 +213,8 @@
 # https://docs.djangoproject.com/en/3.2/howto/static-files/
 # TODO enable static file hosting
 
-# STATIC_URL = '/static/'
-# STATIC_ROOT = os.path.join(BASE_DIR, "static")
-# STATICFILES_DIRS = ((os.path.join(STATIC_ROOT, "static")),)
+STATIC_URL = '/static/'
+STATIC_ROOT = os.path.join(BASE_DIR, "static")
```
caseycesari, line 217: When visiting the site, you should see a demo React app and not the blank screen (you can see what the demo app looks like by visiting the webpack dev server at http://localhost:4545/). The blank screen indicated that some JS files weren't being served properly. If you include the `STATICFILES_DIR` line that was commented out on line 221, it allows the JS files to be loaded properly and for the React app to load from http://localhost:8181/

src/django/iow/settings.py
```diff
@@ -216,9 +213,8 @@
 # https://docs.djangoproject.com/en/3.2/howto/static-files/
 # TODO enable static file hosting
```
caseycesari, line 214: Stale comment

caseycesari (approved):
Nice work! Just one quick change to get this working, and a little bit of clean-up.

src/django/iow/settings.py
nanotubing: Thanks for catching this. I couldn't figure out if this was necessary or not, and after checking a few different projects it looked like some did and some did not have it, so I figured I'd try without it. Lesson learned.

Issue #17, opened 2022-08-08
Upgrade Node to 18


rajadain:
## Overview

We are currently using Node 16 which is nearing end of life in a year. Node 18 will give us longer to work with.

caseycesari:
FYI, here is a commit that removed [hot reload](https://github.com/azavea/pwd-stormwater-allocation/pull/2213/files) on another project, for reference.

PR #19, opened 2022-08-08
Upgrade to Node 18

mstone121:
## Overview

Node 16 will soon be out-of-date as Node 18 is set to become the LTS version in October 2022. 

Connects #17 

### Notes

The only issue with upgrading has been a cryptic error that seems to be related to openssl:

```
Error: error:0308010C:digital envelope routines::unsupported
    at String.replace (<anonymous>)
```

Forcing node to use a legacy ssl provider is a temporary workaround. More info from can be found [in the node release notes](https://nodejs.org/en/blog/release/v17.0.0/#openssl-3-0).

I think the root cause is a module that's using unsupported openssl functions. If there is a way to get webpack to output more logging info while it's bundling, it could help track down which module it might be coming from.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,6 +22,7 @@ services:
       - CHOKIDAR_USEPOLLING=true
       - CHOKIDAR_INTERVAL=100
       - PORT=4545
+      - NODE_OPTIONS=--openssl-legacy-provider
     volumes:
         - ./src/app:/usr/local/src
         - ./src/django/static:/usr/local/src/build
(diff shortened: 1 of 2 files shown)


rajadain (approved):
+1 tested! This is working well. While the legacy SSL option doesn't sound the _most_ secure, it is no worse off than Node 16, and should be fine for the development environment in which we use it.

Nice clean diff!

rajadain:
Reminder to add a CHANGELOG entry before merging!

PR #46, opened 2022-08-25
Reference Image Layer

### LINKED ISSUE #42: Add sample reference layer
Add a sample reference layer that appears in the sidebar under the "Reference layers" heading, and name it something like "Neighborhood map".

While we won't support uploading in the prototype phase, this is to demonstrate the type of image layer that could be added by a user for tracing. Consider using this [sample image](https://upload.wikimedia.org/wikipedia/commons/thumb/2/2a/Sanborn_Fire_Insurance_Map_from_Newark%2C_Essex_County%2C_New_Jersey._LOC_sanborn05571_001-30.tif/lossy-page1-652px-Sanborn_Fire_Insurance_Map_from_Newark%2C_Essex_County%2C_New_Jersey._LOC_sanborn05571_001-30.tif.jpg).

They layer should be toggable like the basemap layers, but should be loaded as a [distortable image layer](https://github.com/publiclab/Leaflet.DistortableImage).

mstone121:
## Overview

This PR add the reference image layer to the draw page. The layer is based on [Leaflet.DistortableImage](https://github.com/publiclab/Leaflet.DistortableImage). 

Closes #42 

### Demo

https://user-images.githubusercontent.com/8356789/186715207-8cc6c250-eceb-48b4-8e76-9a68aabf27a2.mov

### Notes

Updating the layer on state change was causing the image to become unselected. A simple work around for this is to remove the state data from the `useEffect` dependencies. This effectively means that the state is stored in two places: the distortable layer's internal state, and the redux global store. This is less than ideal and should be given future consideration.

## Testing Instructions
- Go to draw page
- Reference layer should be visible
- Toggle between each of the layer tools and test their functionality
- Turn off the reference layer and turn it back on, the image should disappear and appear in the same location with the same transformations

## Checklist

- [ ] CI passes after rebase


rajadain:
Taking a look

rajadain (commented):
This looks really good! Great job styling the plugin to look like the app, it fits in great!

Two things before this is ready to merge:

1. Let's move the "Reference image" toggle button from the "Basemap Layers" section to the "Reference Layers" section on the map, and let's rename it to be "Neighborhood map" as specified in #42 
2. Currently, when the reference layer is toggled on and off, although its transformation is preserved, the state of its controls is not. For example, if I'd set it to transparent and locked, then turned the layer off, and turned it back on again, it comes back opaque and set to Distort rather than Locked.

mstone121:
@rajadain  d5ec7e37a8a9dd8ba7b1111bc2adc4f8bf934e28 moves the reference image visibility toggle to the Reference Layers section of the side bar and implements the persistence of the layer mode.

Still not working is the loading of transparency/outline setting from state. I'll create a follow up issue for this.

src/app/src/components/Layers/ReferenceImageLayer.js
```diff
@@ -0,0 +1,83 @@
...
+    const showLayer = useSelector(state => state.map.referenceImage.visible);
+
+    return showLayer ? <ReferenceImageLayer /> : null;
+}
+
+function ReferenceImageLayer() {
+    const dispatch = useDispatch();
+
+    // TODO: Find a way to initialize the image with transparent/outlined enabled
+    const { corners, mode, transparent, outlined } = useSelector(
```
rajadain, line 33: Lint errors may be fixed with:

```suggestion
    const { corners, mode } = useSelector(
```

rajadain (approved):
+1 tested, nice work. Agree with deferring the rest to later.

At first I was surprised that the reference layer is beneath the Parcel and Municipal Boundary layers, but it seems to work. I'd still prefer them to be in the order they show up in the sidebar though, and since now Reference Layer is the first one, it should be on the very top.

Other than that, this is good to go!

rajadain:
Actually wait, I may have found a bug

rajadain:
Never mind the above. The bug may have been the result of a bad merge with #48 which I was using in the demo. This PR itself works correctly.

Once the reference image layer is moved to the very top / front of the map, so that all other layers are beneath it, this can be merged. Currently, it is obscured by the Parcel and Municipal Boundary layers, and cannot be manipulated once under those layers.

PR #58, opened 2022-09-07
Set editing polygon border color by basemap type

### LINKED ISSUE #56: Improve Polygon Boundary for Satellite Basemap
## Overview

Currently, the polygon is hard to see against a dark background such as the satellite basemap:

image

Compare this with how legible it is on a lighter basemap:

image

Make the polygon boundary easier to see on dark basemaps. This may be done using a dual toned border, as described in this wireframe: https://www.figma.com/file/tJPZ0lkT5C3SAyS2g2R1aA/IoW---Boundary-Sync-Tool?node-id=498%3A14829, but which may not be natively supported by Leaflet (investigate).

Alternatively, we could change the color of the border on basemap change.

Figure out the simplest / easiest way to do this and implement.

mstone121:
## Overview

This PR adds code to update the editing polygon's border when the basemap is changed. If the basemap is Satellite, the border will be white, otherwise the border is black. 

Closes #56 

### Demo

Screen Shot 2022-09-07 at 9 23 31 AM

### Notes

Information about the border type could be moved to the basemap components. Then when the basemap components are mounted, they can set the border type in state. The polygon would read this value from state and update accordingly. I didn't do this for this PR because the card asks for the simplest/easiest solution. 

## Testing Instructions

- Go to Draw Page
- Add a polygon
- Change the basemap
- Default and Land and Water basemaps should have the dark border, while satellite should have the light border.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Incorporated changes in app template [#31](https://github.com/azavea/iow-boundary-tool/pull/31)
 - Update styles for Polygon and Municipal Boundary layers [#48](https://github.com/azavea/iow-boundary-tool/pull/48)
 - Update tiles for all basemaps [#48](https://github.com/azavea/iow-boundary-tool/pull/48)
+- Update Edit Polygon border to be lighter on dark backgrounds [#58](https://github.com/azavea/iow-boundary-tool/pull/58)
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


rajadain:
Taking a look at this now.

rajadain (approved):
+1 tested, this is working well:

image

I agree with your note, that would be a better design for when we have multiple basemaps that may need this dark / light treatment. Until then this is good enough. We can refactor if it's warranted at a later date.

PR #59, opened 2022-09-07
Save layer corners on layer remove

### LINKED ISSUE #55: Reference Image Always Added to Center of Map
## Overview

Currently, when an unedited reference image layer is hidden and the map moved, it reinitializes to the center of the map. Once it has been edited, it sticks correctly.

### Expected Behavior

It stays in the initial position.

### Actual Behavior

It reinitializes to the center of the map.

### Steps to Reproduce

- Go to https://staging.iow.azavea.com/draw and hide the reference image
- Move the map
- Show the reference image again
- Observe that it reappears at the center of the map, rather than its original position

### Demo

https://user-images.githubusercontent.com/1430060/187968833-801f1fd5-d0c7-479a-8a39-d4733f649415.mp4

mstone121:
## Overview

This PR adds an event handler to the reference image layer `remove` event to save the layer's corners if they aren't already saved. This allows the position of the layer to be saved, even if it has not been edited. 

Closes #55 

### Demo

https://user-images.githubusercontent.com/8356789/188916595-e9906146-c8e5-4f07-9a0e-2a2750139dca.mov

### Notes

While the add event might make more sense, the corners are not loaded at that time. The reason for the conditional layer._corners check is that corners are not always loaded when the layer remove event is fired. For example, the remove event is fired twice while adding the reference image.

## Testing Instructions

- Go to draw page
- Make note of reference image location, but don't edit it
- Turn off reference image
- Move map
- Turn on reference image
- The reference image should be in the original location (on earth, not the screen)

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fix react-remove-scroll-bar warning [#29](https://github.com/azavea/iow-boundary-tool/pull/29)
+- Don't move reference image when panning map [#59](https://github.com/azavea/iow-boundary-tool/pull/59)
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


rajadain:
Taking a quick look

rajadain (approved):
+1 tested, this is working well! Nice work figuring it out.

Since the logic is unclear and obtuse, I would recommend adding a comment block above the `if` to explain what that does. Just having the same summary as in the notes may suffice.

PR #73, opened 2022-09-19
Add utility model and development users

### LINKED ISSUE #64: Add Utility Model
## Overview

- Should have a PWSID, Name, and Location field
- Add a test Azavea Utility that is centered around Raleigh
- All `contributor` users must belong to a Utility

### LINKED ISSUE #65: Add Default Users in Development
## Overview

- `c1@azavea.com`, associated with the Azavea Utility
- `v1@azavea.com`
- `a1@azavea.com`

The creation of these users should be scripted, and loaded in when setting up the database for development. Consider calling that in `./scripts/setup`.

jacobtylerwalls:
## Overview

- Define the `Utility` model
- Require contributors to be associated with a utility
- Enable editing in the Django admin
- Add a `resetdb` script that populates a test utility and test users
- Default new utilities to Raleigh

Closes #64 
Closes #65

Test users:
a1@azavea.com/password
v1@azavea.com/password
c1@azavea.com/password

### Demo

After running `scripts/setup`, which runs `scripts/resetdb`, you have three development users and a test utility:

three-users-and-utility

raleigh-utility

The utility is required when editing contributors:

requirement

### Notes

- This PR is currently based on #72. When that is merged, I'll rebase on `develop`.
- My understanding of `pwsid` is that it's alphanumeric, see: https://ordspub.epa.gov/ords/sfdw/f?p=108:34:0::NO:34:P34_FISCAL_YEAR,P34_QUARTER:2016,4

## Testing Instructions

- `scripts/update --docker` (or `resetdb`)
- `scripts/server`
- View User edit page in Django admin: http://localhost:8181/admin/api/user and log in as a1@azavea.com/password
  - verify user and utility exist
  - verify user to utility is many to many
  - verify you cannot edit or create contributors without an associated utility
  - verify you can create other users with or without a utility
  - verify only a1@azavea.com is a superuser
- View Utility edit page in Django admin: http://localhost:8181/admin/api/utility
  - verify Azavea Test Utility exists, centered on Raleigh
- Verify `scripts/resetdb` resets your dev database to the default

## Checklist

- [x] CI passes after rebase


rachelekm (commented):
This works and I like the idea of having a `resetdb` script, but I think setting up the dev environment, populating with default users, and resetting the database all have clear individual purposes that would benefit from being more functionally isolated—~~there also may be instances where you may want to re-provision the VM/run updates and not lose your test data along the way~~Edit: Realized this is in `scripts/setup` and not `scripts/update` so there wouldn't be loss save for an unwary developer running an accidental `scripts/setup`. One way to accomplish this might be to instead create default users through fixtures and then document the email/passwords for those default users in the README. For admin users, that fixture could look like:
```
{ 
    "model": "api.User", 
        "pk": 1, 
        "fields": { 
            "email": "a1@azavea.com", 
            "password": “pbkdf2_sha256$260000$JXr2Oyq8ySdIOxyRPu0WmU$0NQ5dviuAESA8Z57JXqNZUQHmt1su2PqYHK4IFtQ7wE=”
            "is_superuser": true, 
            "is_staff": true, 
            "is_active": true
        } 
}
```
I got the hashed password by opening a django shell and running the below:
```
scripts/manage shell
>>> from django.contrib.auth.hashers import make_password 
>>> make_password('password')
```
We could then load these users on setup by adding the following command to the end of `scripts/update` (and the end of `resetdb` as well), which allows us to isolate `scripts/resetdb` a little more and be more consistent with the setup structure of our other projects as well: `./scripts/manage loaddata <fixturename> `
(I think as long as we add a `pk` attributes to our `User` and `Utility` fixtures no duplicates will be added into the database on repeat `scripts/update` calls.)

What do you think?

rachelekm (approved):
Thanks for opening the issue, I think the rest seems well covered!

src/django/setup.cfg
```diff
@@ -0,0 +1,2 @@
+[flake8]
+max-line-length = 88
```
jacobtylerwalls, line 2: @rachelekm We're going with `pyproject.toml` instead, generally?

jacobtylerwalls:
Thanks for the review!

src/django/api/management/commands/resetdb.py
```diff
@@ -0,0 +1,33 @@
...
+
+        # Create test users.
+        User.objects.create_superuser(
+            email="a1@azavea.com",
+            password="password",
+        )
+        User.objects.create_user(
+            email="v1@azavea.com",
+            password="password",
+            role=Role.objects.get(description="VALIDATOR"),
```
rajadain, line 26: Can this use the new `Roles` enum?

src/django/api/models/user.py
```diff
@@ -64,7 +66,20 @@ class User(AbstractBaseUser, PermissionsMixin):
...
+    utilities = models.ManyToManyField(
+        Utility,
+        blank=True,
+        related_name="users",
+    )
+
     def clean(self):
+        if (
+            self.id
+            and self.role.description == "CONTRIBUTOR"
```
rajadain, line 78: Same here

src/django/setup.cfg
rachelekm: If we're anticipating integrating GH Super-Linter support into this project we can define `black` and `flake8` config files as environment values that will override the linters defaults, so no restrictions there! I'll defer to @rajadain if there's a project-specific preference here.

rajadain: No project specific preference. Let's go with the standard.

jacobtylerwalls: @rajadain Over Slack we discussed the possibility of this changing when the Super-Linter is brought into this project, but until then, `echo-locator` for instance used `setup.cfg`: https://github.com/azavea/echo-locator/blob/develop/.github/super-linter.env#L6-L7

jacobtylerwalls: Sorry, missed your replies before posting.

rajadain: Generally prefer consolidating configuration in one place if we can, rather than having both `pyproject.toml` and `setup.cfg`. If both `flake8` and `black` can be configured in one but not the other, let's go with that. If both can be setup in both, than I have a slight preference for `pyproject.toml` which feels more modern.

PR #82, opened 2022-09-29
Create Remaining Initial Django Models

### LINKED ISSUE #69: Design Data Model
## Overview

Study the Figma designs: https://www.figma.com/file/tJPZ0lkT5C3SAyS2g2R1aA/IoW---Boundary-Sync-Tool

Identify entities that encapsulates the submissions, maps, user interactions, and other pieces of information needed for the front-end.

Create Django Models for these entities. Export a visualization of the Django Models for documentation in the repo.

rachelekm:
## Overview

This creates the initial Django models needed for the different states of submission as well as for the starting user interactions on the frontend, as discussed in the issue comments.

Closes #69 

### Demo

An [ADR](doc/arch/adr-001-data-models.md) is available as part of the PR.

Screen Shot 2022-09-28 at 9 20 11 PM

### Notes

- The `State` default instance, North Carolina, is created through a data migration that references a GEOJSON file downloaded during `scripts/fetch-data`. I found the state boundary url [here](https://public.opendatasoft.com/explore/dataset/us-state-boundaries/export/?refine.name=North+Carolina), but not sure if there's a better resource for state boundary geojson files? I also included `scripts/fetch-data` as part of the setup script so that the GEOJSON gets setup before running migrations. I didn't see any downsides to this, but not sure if I'm missing context on fetching the municipality data (the original script purpose)?
- I added `__str__` methods to some of the models where it felt appropriate, but not all. I wasn't sure what language to use to best describe instances like the "approval of the submission of a boundary of a utility" when there wasn't a unique identifier besides the utility, and it seemed to better fit a future task once we better know use case anyways.
- I decided to group together `Submission`, `Approval`, `Review`, and `Annotation` models into one file. These felt like all related pieces of the submission process that could benefit from being organized together, but curious what others think here.
- ~~I made the `max_length` of the `Submission.upload_filename` and `ReferenceImage.filename` fields match the `max_length` of the [Django `FileField`](https://docs.djangoproject.com/en/4.1/ref/models/fields/#filefield) since that seemed like an appropriate reference.~~ These fields' `max_length` are now `255`.

## Testing Instructions

- `./scripts/fetch-data`
- `./scripts/update`
- `./scripts/server`
- Visit the below Django admin pages for each newly created object:
  - [State](http://localhost:8181/admin/api/state/)
  - [Boundary](http://localhost:8181/admin/api/boundary/)
  - [Submission](http://localhost:8181/admin/api/submission/)
  - [ReferenceImage](http://localhost:8181/admin/api/referenceimage/)
  - [Review](http://localhost:8181/admin/api/review/)
  - [Annotation](http://localhost:8181/admin/api/annotation/)
  - [Approval](http://localhost:8181/admin/api/approval/)
- Confirm the correct field and field types for each object
- Confirm correct required field validation on save
- Confirm North Carolina is a created `State` instance and fields are as expected
- Confirm `User` and `Utility` have new `state` fields
- Confirm `Utility`'s `state` field defaults to North Carolina

## Checklist

- [ ] CI passes after rebase


rajadain:
Taking a look

rajadain:
Here's a visualization of this model setup made using:

```bash
./scripts/manage graph_models api > graph.dot
```

And using https://dreampuf.github.io/GraphvizOnline/ to convert that to SVG:

graphviz

scripts/fetch-data
```diff
@@ -41,5 +54,17 @@ if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
...
             fetchData
         fi
+
+        if [[ -f "$DJANGO_DATA_DIR/$DEFAULT_STATE_NC_GEOJSON_URL" ]]; then
+            if [ "${1:-}" = "--force" ]; then
+                echo "Warning: forcing data refetching!"
+                rm "$DJANGO_DATA_DIR/$NC_GEOJSON_FILENAME"
+                fetchDefaultDBData
+            else
+                echo "Skipping default django data fetching: file exists."
```
rajadain, line 64: Nice attention to developer experience here

src/django/api/migrations/0007_add_default_state.py
```diff
@@ -0,0 +1,27 @@
...
+from django.conf import settings
+from django.db import migrations
+from django.contrib.gis.geos import GEOSGeometry
+
+def add_default_state_nc(apps, schema_editor):
+    # Utility relies on a many-to-one relationship with State
+    # We should populate db with a State so we have ability to
+    # add upcoming state field on Utility with default FK id
+    State = apps.get_model('api', 'State')
+    NC_GEOJSON = json.load(open(os.path.join(settings.BASE_DIR, 'data/NC_BOUNDARY.geojson'), 'r'))
```
rajadain, line 14: In general, prefer to open files using `with` so the file is closed after use:

```python
with open(os.path.join(settings.BASE_DIR, 'data/NC_BOUNDARY.geojson'), 'r') as f:
    NC_GEOSJON = json.load(f)
    ...
```

scripts/fetch-data
```diff
@@ -10,13 +10,26 @@ MUNICIPAL_BOUNDARIES_DATA_LOCATION="https://linc.osbm.nc.gov/explore/dataset/mun
 MUNICIPAL_BOUNDARIES_FILENAME=muni.geo.json
 DATA_DIR=src/app/public/data
 
+DEFAULT_STATE_NC_GEOJSON_URL="https://public.opendatasoft.com/explore/dataset/us-state-boundaries/download/?format=geojson&refine.name=North+Carolina&timezone=America/New_York&lang=en"
+NC_GEOJSON_FILENAME=NC_BOUNDARY.geojson
```
jacobtylerwalls, line 14: We should probably .gitignore this.

src/django/api/models/state.py
```diff
@@ -0,0 +1,18 @@
...
+from django.contrib.gis.db import models as gis_models
+
+__all__ = ["State"]
+
+
+class State(models.Model):
+    id = models.CharField(max_length=2, primary_key=True, unique=True)
+    name = models.CharField(max_length=127)
+    boundary = gis_models.PolygonField(geography=True)
+    options = models.JSONField(blank=True, null=True)
```
rajadain, line 11: Instead of `django.db.models.JSONField`, prefer `django.contrib.postgres.fields.JSONField`, to take advantage of Postgres-native JSON support.

rajadain: Also consider adding `help_text` explaining that this JSON dictionary will contain state-specific configuration.

src/django/api/models/state.py
```diff
@@ -0,0 +1,18 @@
+from django.db import models
+from django.contrib.gis.db import models as gis_models
+
+__all__ = ["State"]
+
+
+class State(models.Model):
+    id = models.CharField(max_length=2, primary_key=True, unique=True)
+    name = models.CharField(max_length=127)
+    boundary = gis_models.PolygonField(geography=True)
```
rajadain, line 10: Since "Boundary" is an entity, with "shape" as the polygon field name, let's reuse that here for consistency.

```suggestion
    shape = gis_models.PolygonField(geography=True)
```

src/django/api/models/state.py
```diff
@@ -0,0 +1,18 @@
+from django.db import models
+from django.contrib.gis.db import models as gis_models
+
+__all__ = ["State"]
+
+
+class State(models.Model):
+    id = models.CharField(max_length=2, primary_key=True, unique=True)
```
rajadain, line 8: Prefer using `django.contrib.postgres.fields.CICharField` for a case-insensitive field that will match "NC", "nc", and any other combination.

scripts/fetch-data
rachelekm: Good call!

src/django/api/models/state.py
rajadain: Also, let's use a MultiPolygonField since many states have disconnected polygons in their shape (e.g. Michigan)

src/django/api/models/reference_image.py
```diff
@@ -0,0 +1,18 @@
+from django.db import models
+
+from .boundary import Boundary
+from .user import User
+
+__all__ = ["ReferenceImage"]
+
+
+class ReferenceImage(models.Model):
+    filename = models.CharField(max_length=100, blank=True)
```
rajadain, line 10: Prefer `255` as the filename length which is more standard.

src/django/api/models/reference_image.py
```diff
@@ -0,0 +1,18 @@
...
+    boundary = models.ForeignKey(Boundary, on_delete=models.CASCADE)
```
rajadain, line 11: For `on_delete`, prefer `models.PROTECT` in most cases, except those where accidental deletion of related data is tolerable. I think in this app, we should use `models.PROTECT` in all cases.

src/django/api/models/submission.py
```diff
@@ -0,0 +1,75 @@
...
+
+from .boundary import Boundary
+from .role import Roles
+from .user import User
+
+__all__ = ["Submission", "Review", "Approval", "Annotation"]
+
+
+def limit_by_validator_or_admin():
+    return models.Q(role__pk=Roles.VALIDATOR) | models.Q(role__pk=Roles.ADMINISTRATOR)
```
jacobtylerwalls, line 13: A possibility if you wish to avoid nesting `Q` objects.
```suggestion
    return models.Q(role__pk__in=[Roles.VALIDATOR, Roles.ADMINISTRATOR])
```

src/django/api/models/submission.py
```diff
@@ -0,0 +1,87 @@
...
+    submitted_by = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+        blank=True,
+        null=True,
+        related_name="submitter"
+    )
+    upload_filename = models.CharField(max_length=100, blank=True)
+    upload_edited_at = models.DateTimeField(null=True, blank=True)
+    notes = models.TextField(blank=True)
```
rajadain, line 29: Add `null=True`

rajadain (changes requested):
This looks great! Nice work.

There may be some State design we can borrow from other projects. Only minor comments here and there. Will take another look once addressed.

src/django/api/models/submission.py
rachelekm: I left off `null=True` for the `TextField` and `CharField` so that the value would be an empty string vs NULL, is there a benefit to using NULL here?

src/django/api/models/state.py
rachelekm: Nice, will implement! Will also get rid of the `clean()` method to uppercase that field since that will no longer be needed

src/django/api/models/submission.py
rajadain: NULL is a clearer representation of an empty value. Without it, we'll have to manually specify an empty string when creating a new submission. I'd prefer to have it.

jacobtylerwalls: See updated discussion at: https://azavea.slack.com/archives/C02N5S81A69/p1664466431686839?thread_ts=1664466073.244999&cid=C02N5S81A69

src/django/api/models/state.py
rachelekm: Documenting here since this was a separate slack discussion, the postgres `JSONField` is deprecated and will lose support in Django 4.0 so we'll move forward with the recommended `django.db.models.JSONField`. I'll still update to add `help_text` to this field, though.

src/django/api/migrations/0007_add_default_state.py
```diff
@@ -0,0 +1,30 @@
...
+    NC_STATE.save()
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('api', '0006_create_state'),
+    ]
+
+    operations = [
+        migrations.RunPython(add_default_state_nc),
```
jacobtylerwalls, line 29: Another team discussion topic, but I try to always add a reverse operation for `RunPython`--at the very least [RunPython.noop](https://docs.djangoproject.com/en/3.2/ref/migration-operations/#django.db.migrations.operations.RunPython.noop)--because it makes unapplying migrations easier (which does come up in development, when switching branches).

rachelekm: Would second a team discussion on this one! I popped `RunPython.noop` in for myself when I was reverting some of these migrations for edits but noticed some of our other projects didn't have one so I took it out....but maybe worth bringing back in regardless for our own ease of use?

rajadain:
Taking another look

src/django/api/models/state.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class State(models.Model):
+    id = CICharField(max_length=2, primary_key=True, unique=True)
+    name = models.CharField(max_length=127)
+    shape = gis_models.MultiPolygonField(geography=True)
+    options = models.JSONField(
+        blank=True,
+        null=True,
+        help_text='This JSON dictionary will contain state-specific configuration.',
```
rajadain, line 15:
```suggestion
        help_text='This JSON dictionary contains state-specific configuration.',
```

src/django/api/models/utility.py
```diff
@@ -1,16 +1,25 @@
 from django.db import models
 from django.contrib.gis.db import models as gis_models
 from django.contrib.gis.geos import Point
+from .state import State
 
 RALEIGH = Point(-78.6382, 35.7796)
 
+
+def get_default_state():
+    return State.objects.get(id="NC").id
```
rajadain, line 10: This is tautological, since we're asking for the `id` of the row with `id="NC"`. Let's just do:

```python
    state = models.ForeignKey(
        State, on_delete=models.PROTECT, default="NC"
    )
```

src/django/api/models/submission.py
```diff
@@ -0,0 +1,76 @@
...
+    shape = gis_models.PolygonField(geography=True)
+    created_at = models.DateTimeField(auto_now_add=True)
+    created_by = models.ForeignKey(
+        User, on_delete=models.PROTECT, related_name="creator"
+    )
+    submitted_at = models.DateTimeField(null=True, blank=True)
+    submitted_by = models.ForeignKey(
+        User, on_delete=models.PROTECT, blank=True, null=True, related_name="submitter"
+    )
+    updated_at = models.DateTimeField(null=True, blank=True)
```
rajadain, line 27:
```suggestion
    updated_at = models.DateTimeField(auto_now=True)
```

doc/arch/adr-001-data-models.md
```diff
@@ -0,0 +1,101 @@
...
+    Utility ||--o{ User_Utility : "belongs to"
+    Utility }o--|| State : "within"
+    Utility {
+        string pwsid
+        string name
+        address address
+        point location
+        State state FK
+    }
+    User_Utility {
```
rajadain, line 36: Since `User_Utility` doesn't have any additional columns beyond each's id, we can drop it from here, and just indicate a many-to-many relationship between User and Utility

rajadain (commented):
👏 Excellent work! This is looking great. Should be good to go after a few small tweaks.

rajadain (approved):
+1 tested, this looks great! Nice work carrying this through the different iterations. There may be some more small tweaks in the future, but this solid base sets us up very well. Great job!

PR #83, opened 2022-09-29
Force users to reset password on first login

### LINKED ISSUE #67: Have User Setup New Password on First Login
## Overview

New users will be created by administrators, and will be given a username / password combination. Upon first login, they should see a screen to reset their password, and should not be able to proceed until they have done so.

jacobtylerwalls:
## Overview

Force users to reset their password on first login.

Closes #67 

### Notes
- Using the Django admin to reset a password does not set `has_admin_generated_password` to False, so that flow will not force the user to reset their password. If that is desired, that could be a follow-up card.

## Testing Instructions

- `scripts/resetdb`
- `scripts/server`
- Visit django admin at localhost:8181/admin and add a user
- Login as that user, ensure redirected to Reset password page
- [ ] Reset password and ensure login works
- [ ] Login again, ensure not asked to reset again
- [ ] Login as development users (e.g. v1@azavea.com), ensure that no resetting is required

## Checklist

- [x] CI passes after rebase


src/app/src/constants.js
```diff
@@ -43,5 +43,10 @@ export const API_URLS = {
     LOGIN: 'api/auth/login/',
     LOGOUT: 'api/auth/logout/',
     FORGOT: 'api/auth/password/reset/',
-    RESET: 'api/auth/password/reset/confirm/',
+    CONFIRM: 'api/auth/password/reset/confirm/',
+    RESET: 'confirm_password_reset/reset/',
```
jacobtylerwalls, line 47: This was starting to get confusing, so this is my new proposal (see updates to axios calls, below).

src/django/api/views/auth.py
```diff
@@ -21,6 +24,13 @@ def post(self, request, *args, **kwargs):
 
         if user is None:
             raise AuthenticationFailed("Unable to login with those credentials")
+        if user.has_admin_generated_password:
+            context = {
+                'uid': urlsafe_base64_encode(force_bytes(user.pk)),
+                'token': default_token_generator.make_token(user),
```
jacobtylerwalls, line 30: Reference: https://github.com/django/django/blob/0dd29209091280ccf34e07c9468746c396b7778e/django/contrib/auth/forms.py#L337-L339

rachelekm (approved):
Looks good! Functions as expected and think the naming updates to `API_URLS` makes sense 👍

jacobtylerwalls:
Thanks for the review!

PR #97, opened 2022-09-29
Fix slow performance of Django server in Docker 

jacobtylerwalls:
## Overview
1. Similar to 598a0b3ddc5c49a9aea9001ddf7293bbe926d31e, update a port number to agree with the README (see Line 54 of the README).
2. More substantively, increase the number of `gunicorn` workers to 2, so that requests don't block each other and cause worker timeouts.

## Testing instructions
- `scripts/update --docker`
- `scripts/server`
- Visit the Django admin at localhost:8181/admin (in Chrome, since the error was not reproducible for me in Safari), and visit multiple pages: ensure app does not lock up.

## Notes
See similar approach here:
https://github.com/azavea/iow-boundary-tool/blob/2b3f90bbacd1377ee2734b6da557c51cb8e50a6a/src/django/Dockerfile#L28

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,14 @@ vagrant ssh
 ./scripts/server
 ```
 
+After running `setup` (or `resetdb`), three test users are created:
+
+| User                 | Password          | Role          |
+| ---------------------|-------------------|---------------|
+| a1@azavea.com        | password          | administrator |
+| v1@azavea.com        | password          | validator     |
+| c1@azavea.com        | password          | contributor   |
+
 ### Ports
 
 | Service            | Port                            |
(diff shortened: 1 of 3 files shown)


jacobtylerwalls:
@JN-Hernandez you are of course welcome to take a look, but the second commit ended up fixing the issue @rachelekm noticed when working with the Django admin, so if you don't mind I'm going to request/assign her to see if this solves it for both of us!

rachelekm (approved):
Amazing! So much faster on my end 👍

JN-Hernandez:
@jacobtylerwalls Don't mind at all!

JN-Hernandez:
Per our slack discussion, we might need to update the README regarding where to find test credentials.  Other than that, I was able to log in and navigate multiple pages without the application locking up.  +1

PR #101, opened 2022-10-03
Add RTK Query

### LINKED ISSUE #86: Add RTK Query
## Overview

We need a uniform strategy / toolset for API interaction that integrates with Redux. We've used manual [axios](https://axios-http.com/) calls in the past, sometimes combined with [SWR](https://swr.vercel.app/) for reads.

[RTK Query](https://redux-toolkit.js.org/rtk-query/overview) is a similar tool, but one that integrates well with Redux Toolkit, which we're [already using](https://github.com/azavea/iow-boundary-tool/blob/faaede8e49fdb5b4bd294a24020d887dd4199b7c/src/app/package.json#L20) in this app.

Import and add scaffolding for RTK Query, and switch any applicable existing interactions (Login / Forgot Password) to it.

mstone121:
## Overview

This PR adds RTK Query endpoints. RTK Query will be used to access to majority of the API.

Closes #86 

## Testing Instructions

- The app should build and load

As of this PR, none of the components are access the API (besides the auth components which don't fit into the RTK Query query/mutation pattern very well). The endpoints in this PR are not real and are there for now as a reference.

## Notes

I moved `/src/api.js` to `/src/api/client.js` and updated the name to `apiClient` where it was used. This should help to distinguish it from the RTK Query "api".

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/api/api.js
+++ b/src/app/src/api/api.js
@@ -1,0 +1,28 @@
+import { createApi } from '@reduxjs/toolkit/query/react';
+
+import apiClient from './client';
+import { BASE_API_URL } from '../constants';
+
+import TAGS from './tags';
+
+const axiosBaseQuery = ({ url, method, data, params }) =>
+    apiClient({
+        url: `${BASE_API_URL}${url}`,
+        method,
+        data,
+        params,
+    })
+        .then(result => ({ data: result.data }))
+        .catch(error => ({
+            error: {
+                status: error.response?.status,
+                data: error.response?.data || error.message,
+            },
+        }));
+
+export default createApi({
+    reducerPath: 'api',
+    baseQuery: axiosBaseQuery,
+    tagTypes: Object.values(TAGS),
+    endpoints: () => ({}),
+});
--- a/src/app/src/api/submissions.js
+++ b/src/app/src/api/submissions.js
@@ -1,0 +1,52 @@
+import api from './api';
+import TAGS, {
+    getListTagProvider,
+    getNewItemTagInvalidator,
+    getSingleItemProvider,
+    getUpdateItemTagInvalidator,
+} from './tags';
+
+const submissionApi = api.injectEndpoints({
+    endpoints: build => ({
+        getDraftSubmissions: build.query({
+            query: () => ({
+                url: '/submissions',
+                method: 'GET',
+                params: { type: 'drafts' },
+            }),
+            providesTags: getListTagProvider(TAGS.SUBMISSION),
+        }),
+
+        getSubmissionDetails: build.query({
+            query: id => ({
+                url: `/submissions/${id}`,
+                method: 'GET',
+            }),
+            providesTags: getSingleItemProvider(TAGS.SUBMISSION),
+        }),
+
+        startNewSubmission: build.mutation({
+            query: newSubmission => ({
+                url: '/submissions',
+                method: 'POST',
+                body: newSubmission,
+            }),
+            invalidatesTags: getNewItemTagInvalidator(TAGS.SUBMISSION),
+        }),
+
+        submitSubmission: build.mutation({
+            query: id => ({
+                url: `/submissions/submit/${id}`,
+                method: 'POST',
+            }),
+            invalidatesTags: getUpdateItemTagInvalidator(TAGS.SUBMISSION),
+        }),
+    }),
+});
+
+export const {
+    useGetDraftSubmissionsQuery,
+    useGetSubmissionDetailsQuery,
+    useStartNewSubmissionMutation,
+    useSubmitSubmissionMutation,
+} = submissionApi;
--- a/src/app/src/api/tags.js
+++ b/src/app/src/api/tags.js
@@ -1,0 +1,26 @@
+const TAGS = {
+    SUBMISSION: 'submission',
+
+    LIST_ID: 'list',
+};
+
+export function getListTagProvider(tag) {
+    return results => [
+        ...results.map(({ id }) => ({ tag, id })),
+        { tag, id: TAGS.LIST_ID },
+    ];
+}
+
+export function getSingleItemProvider(tag) {
+    return (result, error, id) => ({ tag, id });
+}
+
+export function getNewItemTagInvalidator(tag) {
+    return () => ({ tag, id: TAGS.LIST_ID });
+}
+
+export function getUpdateItemTagInvalidator(tag) {
+    return (result, error, id) => ({ tag, id });
+}
+
+export default TAGS;
--- a/src/app/src/store/store.js
+++ b/src/app/src/store/store.js
@@ -1,11 +1,15 @@
 import { configureStore } from '@reduxjs/toolkit';
 
+import api from '../api';
 import authReducer from './authSlice';
 import mapReducer from './mapSlice';
 
 export default configureStore({
     reducer: {
+        [api.reducerPath]: api.reducer,
         auth: authReducer,
         map: mapReducer,
     },
+    middleware: getDefaultMiddleware =>
+        getDefaultMiddleware().concat(api.middleware),
 });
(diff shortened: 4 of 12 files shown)


src/app/src/api/api.js
```diff
@@ -0,0 +1,33 @@
...
+        .catch(error => {
+            dispatch(setApiError(error.message));
+
+            return {
+                error: {
+                    status: error.response?.status,
+                    data: error.respone?.data || error.message,
+                },
+            };
+        });
```
mstone121, line 26: These return types are expected by RTK Query:
```js
// Success
{ data: YourData }

// Error
{ error: YourError }
```
https://redux-toolkit.js.org/rtk-query/usage/customizing-queries#implementing-a-custom-basequery

src/app/src/api/submissions.js
```diff
@@ -0,0 +1,48 @@
...
+    getUpdateItemTagInvalidator,
+} from './tags';
+
+const submissionApi = api.injectEndpoints({
+    endpoints: build => ({
+        getDraftSubmissions: build.query({
+            query: () => ({
+                url: '/submissions',
+                method: 'GET',
+                params: { type: 'drafts' },
```
mstone121, line 15: Probably not the best URL scheme: `/submissions?type=drafts`, but I included it to demonstrate how query parameters can be used.

src/app/src/api/tags.js
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export function getNewItemTagInvalidator(tag) {
+    return () => ({ tag, id: TAGS.LIST_ID });
+}
+
+export function getUpdateItemTagInvalidator(tag) {
+    return (result, error, id) => ({ tag, id });
+}
+
```
mstone121, line 25: In my past experience with RTK Query, managing the API tags for cache invalidation can be a pain point. I added these helper functions mostly as a way of guiding future implementation of tags. We needn't get too attached to this pattern though. It might turn out they are not flexible enough for some parts of our data schema. 

Tags are the recommended way to go from RTK Query, but it is possible to manually update the cache. More info can be found here: https://redux-toolkit.js.org/rtk-query/usage/manual-cache-updates.

src/app/src/api/submissions.js
```diff
@@ -0,0 +1,48 @@
+import api from './api';
```
mstone121, line 1: This file is just here for reference at the moment. It does not reflect actual API routes.

rajadain:
Taking a look now

src/app/src/store/store.js
```diff
@@ -1,11 +1,15 @@
 import { configureStore } from '@reduxjs/toolkit';
 
+import api from '../api';
 import authReducer from './authSlice';
 import mapReducer from './mapSlice';
```
rajadain, line 5: Should we also be importing `errorSlice` here?

src/app/src/store/store.js
```diff
@@ -1,11 +1,15 @@
 import { configureStore } from '@reduxjs/toolkit';
 
+import api from '../api';
```
rajadain, line 3: Since `api/api` is importing from `store/errorSlice`, and we're importing `api/api` into `store/store`, this makes accidental circular references more likely in the future.

Should the RTK Query content be moved into `store`? Is there a better way to organize this?

src/app/src/api/submissions.js
```diff
@@ -0,0 +1,52 @@
...
+                url: `/submissions/submit/${id}`,
+                method: 'POST',
+            }),
+            invalidatesTags: getUpdateItemTagInvalidator(TAGS.SUBMISSION),
+        }),
+    }),
+});
+
+export const {
+    useGetDraftSubmissionsQuery,
```
rajadain, line 48: Would be curious for a brief demo of how this hook gets used

rajadain (approved):
+1 tested. This looks good. I'm not familiar with RTK Query, but cautiously optimistic for trying it out. Some of these patterns will become clearer as we start to fetch data.

src/app/src/api/submissions.js
mstone121: [`useMutation`](https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#usemutation) is slightly different in it's usage:

```jsx
function NewDraftPage() {
    const [formValues, setFormValues] = useState({});
    const [createNewDraft, { isLoading }] = useCreateNewDraftMutation();

    function handleCreateNewDraft() {
        createNewDraft(formValues);
    }

   if (isLoading) {
        return <CircularProgress />
    } 

    return (<Form>
        <Button onClick={handleCreateNewDraft}>Create New Draft</Button>
    </Form>);
}

File: src/app/src/store/store.js (hunk 5 above)
Comment (mstone121) [reply src/app/src/store/store.js:5]: 15f1ffefe859b48e89152a38e56ce203218b8fb9

File: src/app/src/store/store.js (hunk 6 above)
Comment (mstone121) [reply src/app/src/store/store.js:3]: I'm not sure. I like having a top level folder for the api because it will likely accumulate a number of files and they will be referenced from many different places. On the other hand, the endpoints/api are a part of the store and so logically they fit in that folder.

**Comment by rajadain** [comment]
Since in this architecture, there is only one `errorSlice`, do different endpoints overwrite their error responses?

I've seen Redux response architectures like this:

```js
{
  submissions: {
    data: null,
    isFetching: false,
    error: null,
  },
  user: {
    data: null,
    isFetching: false,
    error: null,
  }
}
```

In this case, the `submissions` and `user` keys have their own `error` record. In the above, we're using a top level key, more like this:

```js
{
  submissions: {
    data: null,
    isFetching: false,
  },
  user: {
    data: null,
    isFetching: false,
  },
  error: {
    apiError: null,
  }
}
```

Is this understanding correct?

If RTK Query gives us per-endpoint `error` out of the box as demo'd here:

```js
const { isFetching, data: drafts, error } = useGetDraftSubmissionQuery(
    // If we need to pass args to the endpoint, they would go here.
    // For example, if this were useGetSubmissionDetails, the id would go here.
);
```

Do we need `errorSlice`?

mstone121:
> Do we need `errorSlice`?

It is very possible that we do not. The idea was to have some sort of global error handling so we're not re-creating an error message handler for every component that uses an endpoint:
```jsx
<App>
    <ErrorDialog show={errorSlice.apiError !== null} />
    ... rest of the app ...
```
This might not be the best idea though. It won't handle multiple errors and perhaps we want some errors to fail silently. I'm good with removing `errorSlice` for now and implementing something later after we've implemented more components.

rajadain:
> I'm good with removing `errorSlice` for now

I would recommend this. Let's handle errors within components for now, and only reach for app-wide error handling when we get there.

src/app/src/api/api.js
```diff
@@ -0,0 +1,28 @@
...
+        url: `${BASE_API_URL}${url}`,
+        method,
+        data,
+        params,
+    })
+        .then(result => ({ data: result.data }))
+        .catch(error => ({
+            error: {
+                status: error.response?.status,
+                data: error.respone?.data || error.message,
```
jacobtylerwalls, line 19:
```suggestion
                data: error.response?.data || error.message,
```

mstone121: Whoops, good catch. [29cfe14](https://github.com/azavea/iow-boundary-tool/pull/101/commits/29cfe144371a6e294279ff9b20f1c2079864b29b)

jacobtylerwalls (approved):
Thanks for letting me have a look. I understand this at a high-level and am looking forward to seeing the first implementation!

src/app/src/api/api.js
```diff
@@ -0,0 +1,28 @@
...
+        .then(result => ({ data: result.data }))
+        .catch(error => ({
+            error: {
+                status: error.response?.status,
+                data: error.response?.data || error.message,
+            },
+        }));
+
+export default createApi({
+    reducerPath: 'api',
```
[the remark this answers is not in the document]
mstone121: This is the path of `api` in the reducer, not anything with the external API.

I'm pretty sure it just sets a property of the api object which is [later picked up in the store](https://github.com/azavea/iow-boundary-tool/pull/101/files#diff-339d12e291cd070d889c443e066d72e3481236aadac73f382644e4935a165ee4R9).

src/app/src/store/store.js
rajadain: Alright, let's keep it for now, but also keep an eye on this.

rajadain (approved):
This looks good to me! Some new patterns, but enough overlap with SWR in use to be familiar. Nice job setting this up, and thanks for providing the demos for use!

PR #111, opened 2022-10-06
Add State Boundaries

### LINKED ISSUE #84: Populate All States
## Overview

Use the TIGER/LINE database: https://www2.census.gov/geo/tiger/GENZ2021/shp/cb_2021_us_state_20m.zip, as done [in other projects](https://azavea.slack.com/archives/C02N5S81A69/p1664480687915709).

Convert the above Shapefile to GeoJSON and write a migration to populate all States. Since this data is static, it can be done as a migration.

mstone121:
## Overview

This PR updates `./scripts/fetch-data` to fetch US State boundaries from the TIGER/LINE database. 

This data is then used in a data migration to add the state boundaries to the database.

Closes #84 

### Notes

I added a `--dev` flag to `./scripts/fetch-data` since the state data won't be needed on every build, just when the data migration is run. It seems wasteful to download this resource on every build when it is needed only once.

GDAL must be installed for the fetch data script to work.

## Testing Instructions

* `./scripts/fetch-data --dev`
* `./scripts/manage migrate`

## Checklist

- [x] CI passes after rebase


mstone121:
I've realized that since the build will migrate the database from scratch, it will be required on each build. The automated-fetching process also converts the data from shp to geojson which requires GDAL, but this appears to be installed in the django container.

I'm not certain what to do at this point. Perhaps it would be better to commit the geojson file. It would still be easy to update if necessary.

rajadain:
> Perhaps it would be better to commit the geojson file. It would still be easy to update if necessary.

I would be in favor of committing the GeoJSON file, and using the fetch-data script to update it when we need to.

rajadain:
> The automated-fetching process also converts the data from shp to geojson which requires GDAL, but this appears to be installed in the django container.

Can we rewrite that bit to be done within the Django container then? So the developer doesn't need to install GDAL on their machine just for this?

rajadain:
Also, we would need these states populated in Staging and Production as well, so everything the migration needs should be in the deployed artifact (easiest to do by committing it to the repo).

We should be able to run:

```
./scripts/manage ecsmanage migrate
```

and have the boundaries be correctly installed on staging.

mstone121:
@rajadain This is ready for review again. I moved the `ogr2ogr` part of the script into the container so the user doesn't need to install GDAL: https://github.com/azavea/iow-boundary-tool/pull/111/commits/525bae99b36f829dd8fd254b1619cbdb97c2e95f

I added the geojson file to VCS with https://github.com/azavea/iow-boundary-tool/pull/111/commits/5caf68427f52f7b124675c45eecabbabfcd2577b.

rajadain:
Taking another look

scripts/fetch-data
```diff
@@ -6,65 +6,109 @@ if [[ -n "${IOW_BOUNDARY_TOOL_DEBUG}" ]]; then
...
+
+function unzipStateBoundaryData() {
+    rm -rf "${DJANGO_DATA_DIR:?}/$STATE_BOUNDARIES_SHAPE_DIR"
+    unzip "$DJANGO_DATA_DIR/$STATE_BOUNDARIES_ZIP_FILENAME" \
+        -d "$DJANGO_DATA_DIR/$STATE_BOUNDARIES_SHAPE_DIR"
+
+    rm -f "$DJANGO_DATA_DIR/$STATE_BOUNDARIES_FILENAME"
+
+    # Run ogr2ogr in docker container
+    docker-compose exec django ogr2ogr -f geojson \
```
rajadain, line 91: This should not rely on a running container

```suggestion
    docker-compose run --rm django ogr2ogr -f geojson \
```

rajadain (approved):
+1 tested, nicely done! This works well, and has all the data we need.

PR #113, opened 2022-10-07
Add boundary views and serializers

### LINKED ISSUE #107: Add Serializers, Views for Submissions
## Overview

Add serializers and views for fetching a list of submissions and an individual submission. Come up with data structures for each.

The list endpoint should be minimal, with just enough information to populate the Submission List view. It should support filtering by utility.

The detail endpoint should fetch all information needed for the Submission Detail view.

~~The endpoints should generally respond in a shape like:~~

```json
{ "data": {}, "error": {} }
```

~~to work well with RTK Query.~~

Add security checks that allow Validators of the right State to access this.

Contributors should only be able to access submissions that belong to a Utility that they have access to.

mstone121:
## Overview

This PR adds serializers for submissions and a few related objects and views for submissions.

Closes #107 

### Notes

I added a venv folder to the .gitignore so I could install requirements locally.

TODO: Create activity log

## Testing Instructions

* `scripts/manage resetdb` in order to test the validator user, if desired
* Log in as c1@azavea.com
* Go to http://localhost:8181/api/boundaries/
* See list of submissions
  * There should be one for each stage in the submission process
* Pick one of the submissions, go to the details page (i.e. http://localhost:8181/api/boundaries/1)
  * Check that all the relevant details are returned (reviews/annotations/submissions)

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ deployment/ansible/roles/azavea.*
 # Django
 /src/django/static/
 /src/django/data/
+/.venv
 
 # JS
 node_modules/
--- a/src/django/api/management/commands/resetdb.py
+++ b/src/django/api/management/commands/resetdb.py
@@ -6,13 +6,14 @@ from django.core.management.base import BaseCommand
 from api.models import Utility, User
 from api.models.boundary import Boundary
 from api.models.user import Roles
-from api.models.submission import Approval, Submission
 from api.models.state import State
+from api.models.submission import Approval, Submission, Review, Annotation
 
 from ..test_shapes import (
     RALEIGH_FAKE_RECTANGLE,
     RALEIGH_FAKE_TRIANGLE,
     RALEIGH_FAKE_ZIGZAG,
+    POINT_IN_RALEIGH_FAKE_TRIANGLE,
 )
 
 
@@ -23,7 +24,7 @@ class Command(BaseCommand):
         call_command("reset_schema", "--noinput")
         call_command("migrate")
 
-        # Create test utility located in Raleigh.
+        # Create test utilities located in Raleigh.
         test_utility = Utility(pwsid="123456789", name="Azavea Test Utility")
         test_utility.save()
 
@@ -48,15 +49,15 @@ class Command(BaseCommand):
             has_admin_generated_password=False,
             role=Roles.CONTRIBUTOR,
         )
-        contributor.utilities.add(test_utility_1)
-        contributor.utilities.add(test_utility_2)
-        contributor.utilities.add(test_utility_3)
+        contributor.utilities.add(test_utility)
 
         # Create test boundaries and submissions.
         # Use tz-aware datetimes to avoid warnings.
-        boundary_1 = Boundary.objects.create(utility=test_utility_1)
-        boundary_2 = Boundary.objects.create(utility=test_utility_2)
-        boundary_3 = Boundary.objects.create(utility=test_utility_3)
+        boundary_1 = Boundary.objects.create(utility=test_utility)
+        boundary_2 = Boundary.objects.create(utility=test_utility)
+        boundary_3 = Boundary.objects.create(utility=test_utility)
+        boundary_4 = Boundary.objects.create(utility=test_utility)
+        boundary_5 = Boundary.objects.create(utility=test_utility)
 
         # draft
         Submission.objects.create(
@@ -77,8 +78,60 @@ class Command(BaseCommand):
             notes="Notes for the test submission.",
         )
 
-        approved = Submission.objects.create(
+        # reviewing
+        submitted = Submission.objects.create(
             boundary=boundary_3,
+            shape=RALEIGH_FAKE_TRIANGLE,
+            created_at=datetime(2022, 10, 2, hour=4, tzinfo=timezone.utc),
+            created_by=contributor,
+            submitted_at=datetime(2022, 10, 2, hour=6, tzinfo=timezone.utc),
+            submitted_by=contributor,
+            notes="Notes for the test submission.",
+        )
+
+        review = Review.objects.create(
+            submission=submitted,
+            reviewed_by=validator,
+            notes="Notes for the review.",
+            created_at=datetime(2022, 10, 2, hour=10, tzinfo=timezone.utc),
+        )
+
+        Annotation.objects.create(
+            review=review,
+            location=POINT_IN_RALEIGH_FAKE_TRIANGLE,
+            comment="Comment on review",
+            created_at=datetime(2022, 10, 2, hour=10, minute=5, tzinfo=timezone.utc),
+        )
+
+        # needs revision
+        submitted = Submission.objects.create(
+            boundary=boundary_4,
+            shape=RALEIGH_FAKE_TRIANGLE,
+            created_at=datetime(2022, 10, 2, hour=4, tzinfo=timezone.utc),
+            created_by=contributor,
+            submitted_at=datetime(2022, 10, 2, hour=6, tzinfo=timezone.utc),
+            submitted_by=contributor,
+            notes="Notes for the test submission.",
+        )
+
+        review = Review.objects.create(
+            submission=submitted,
+            reviewed_by=validator,
+            reviewed_at=datetime(2022, 10, 2, hour=11, tzinfo=timezone.utc),
+            notes="Notes for the review.",
+            created_at=datetime(2022, 10, 2, hour=10, tzinfo=timezone.utc),
+        )
+
+        Annotation.objects.create(
+            review=review,
+            location=POINT_IN_RALEIGH_FAKE_TRIANGLE,
+            comment="Comment on review",
+            created_at=datetime(2022, 10, 2, hour=10, minute=5, tzinfo=timezone.utc),
+        )
+
+        # approved
+        approved = Submission.objects.create(
+            boundary=boundary_5,
             shape=RALEIGH_FAKE_RECTANGLE,
             created_at=datetime(2022, 10, 1, hour=8, tzinfo=timezone.utc),
             created_by=contributor,
--- a/src/django/api/management/commands/resetdb.py
+++ b/src/django/api/management/commands/resetdb.py
@@ -7,6 +7,7 @@ from api.models import Utility, User
 from api.models.boundary import Boundary
 from api.models.user import Roles
 from api.models.submission import Approval, Submission
+from api.models.state import State
 
 from ..test_shapes import (
     RALEIGH_FAKE_RECTANGLE,
@@ -32,27 +33,34 @@ class Command(BaseCommand):
             password="password",
             has_admin_generated_password=False,
         )
+
         validator = User.objects.create_user(
             email="v1@azavea.com",
             password="password",
             has_admin_generated_password=False,
             role=Roles.VALIDATOR,
         )
+        validator.states.add(State.objects.get(pk='NC'))
+
         contributor = User.objects.create_user(
             email="c1@azavea.com",
             password="password",
             has_admin_generated_password=False,
             role=Roles.CONTRIBUTOR,
         )
-        contributor.utilities.add(test_utility)
+        contributor.utilities.add(test_utility_1)
+        contributor.utilities.add(test_utility_2)
+        contributor.utilities.add(test_utility_3)
 
-        # Create test boundary and submissions.
+        # Create test boundaries and submissions.
         # Use tz-aware datetimes to avoid warnings.
-        boundary = Boundary.objects.create(utility=test_utility)
+        boundary_1 = Boundary.objects.create(utility=test_utility_1)
+        boundary_2 = Boundary.objects.create(utility=test_utility_2)
+        boundary_3 = Boundary.objects.create(utility=test_utility_3)
 
         # draft
         Submission.objects.create(
-            boundary=boundary,
+            boundary=boundary_1,
             shape=RALEIGH_FAKE_ZIGZAG,
             created_at=datetime(2022, 10, 3, hour=15, tzinfo=timezone.utc),
             created_by=contributor,
@@ -60,7 +68,7 @@ class Command(BaseCommand):
 
         # submitted
         Submission.objects.create(
-            boundary=boundary,
+            boundary=boundary_2,
             shape=RALEIGH_FAKE_TRIANGLE,
             created_at=datetime(2022, 10, 2, hour=11, tzinfo=timezone.utc),
             created_by=contributor,
@@ -70,7 +78,7 @@ class Command(BaseCommand):
         )
 
         approved = Submission.objects.create(
-            boundary=boundary,
+            boundary=boundary_3,
             shape=RALEIGH_FAKE_RECTANGLE,
             created_at=datetime(2022, 10, 1, hour=8, tzinfo=timezone.utc),
             created_by=contributor,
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -21,6 +21,7 @@ deployment/ansible/roles/azavea.*
 # Django
 /src/django/static/
 /src/django/data/
+/venv
```
rajadain, line 24: Minor, but I use a hidden `.venv` since the folder rarely needs to be looked in to.

I also have `.venv` ignored in my global `~/.gitignore`, but adding it here makes sense too.

src/django/api/serializers/submission.py
```diff
@@ -0,0 +1,22 @@
+from rest_framework.serializers import ModelSerializer, CharField
+
+from .route_based_serializer_mixin import RouteBasedSerializerMixin
```
rajadain, line 3: It's unclear what this does

src/django/api/views/submission.py
```diff
@@ -0,0 +1,20 @@
...
+
+from ..models.submission import Submission
+from ..serializers.submission import SubmissionSerializer
+from ..filters.submission import (
+    UserSubmissionFilterBackend,
+    RequestSubmissionFilterBackend,
+)
+
+
+class SubmissionsViewSet(ListModelMixin, GenericViewSet):
```
rajadain, line 13: I've sometimes had trouble making custom actions for ViewSets in the past, so have preferred simple Views, but am willing to give this a shot, in case things have gotten better.

They do document a way to have custom actions: https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing

For if we want something like:

```
/api/submissions/42/approve/
```

src/django/api/views/submission.py
```diff
@@ -0,0 +1,20 @@
...
+    queryset = Submission.objects.all()
+    serializer_class = SubmissionSerializer
+    permission_classes = [IsAuthenticated]
+    filter_backends = [
+        UserSubmissionFilterBackend,
+        RequestSubmissionFilterBackend,
```
rajadain, line 19: Consider using simpler names that communicate what this does, rather than what it is:

```suggestion
        ByRole,
        ByUtility,
```

src/django/api/serializers/submission.py
rajadain: Ah is this for having different fields when pulling the List view vs Detail view? In this case, prefer having two separate serializer classes for the List and Detail views, and using `get_serializer_class()` in `SubmissionsViewSet` to pick the right one dynamically. That would clearer than the metaprogramming the in mixin.

rajadain (commented):
Left some comments.

My tendency has been to used function-based views in the past rather than class-based ones. ViewSets add standardization and offer good built-in support, but can make custom / non-standard work a little difficult. DRF has a brief write-up on this: https://www.django-rest-framework.org/tutorial/6-viewsets-and-routers/#trade-offs-between-views-vs-viewsets

For now, I think ViewSets are fine. If we need to refactor to function-based views down the line, we can do it then.

src/django/api/serializers/submission.py
rajadain: Example: https://stackoverflow.com/a/22755648/6995854

src/django/api/filters/submission.py
```diff
@@ -0,0 +1,28 @@
+from rest_framework.filters import BaseFilterBackend
+
+from ..models.role import Roles
+
+
+class UserSubmissionFilterBackend(BaseFilterBackend):
+    def filter_queryset(self, request, queryset, view):
+        if request.user.role == Roles.CONTRIBUTOR:
```
jacobtylerwalls, line 8: I just ran into the following problem myself, so I assume it applies here. (I'm thinking I should have gone with my original idea of making the `description` field the pk for `Role`.)

***
```python
>>> u.role
<Role: CONTRIBUTOR>
>>> u.role == Roles.CONTRIBUTOR
False
>>> u.role.pk == Roles.CONTRIBUTOR
True
```

jacobtylerwalls: Could add something like this:
```diff
diff --git a/src/django/api/models/role.py b/src/django/api/models/role.py
index b532118..0ddf988 100644
--- a/src/django/api/models/role.py
+++ b/src/django/api/models/role.py
@@ -23,3 +23,6 @@ class Role(models.Model):
 
     def __str__(self):
         return self.description
+
+    def __eq__(self, other):
+        return self.pk.__eq__(other)
```

src/django/api/filters/submission.py
```diff
@@ -0,0 +1,28 @@
...
+            queryset.filter(boundary__utility__in=request.user.utilities)
```
jacobtylerwalls, line 9: Sorry for the piecemeal review. Other things I noticed while working on the ReferenceImage serializer.

m2m relationships need all() to get a queryset, otherwise `.utilities` is just a manager (and isn't iterable)
```suggestion
            queryset = queryset.filter(boundary__utility__in=request.user.utilities.all())
```

src/django/api/filters/submission.py
```diff
@@ -0,0 +1,28 @@
...
+
+        if request.user.role == Roles.VALIDATOR:
+            queryset.filter(boundary__utility__state__in=request.user.states)
```
jacobtylerwalls, line 12:
```suggestion
            queryset = queryset.filter(boundary__utility__state__in=request.user.states.all())
```

src/django/api/filters/submission.py
```diff
@@ -0,0 +1,28 @@
...
+
+        return queryset
+
+
+class RequestSubmissionFilterBackend(BaseFilterBackend):
+    def filter_queryset(self, request, queryset, view):
+        requested_utilities = request.data.get('utilities')
+
+        if requested_utilities:
+            queryset.filter(utility__in=requested_utilities)
```
jacobtylerwalls, line 26: .filter() doesn't operate in place.
```suggestion
            queryset = queryset.filter(utility__in=requested_utilities)
```

jacobtylerwalls (commented):
Sorry for the piecemeal review. Other things I noticed while working on the ReferenceImage serializer.

src/django/api/filters/submission.py
mstone121: https://github.com/azavea/iow-boundary-tool/pull/117

src/django/api/serializers/submission.py
mstone121: Cool. I was looking for a way to do exactly that, but could not find it in the documentation.

src/django/api/views/submission.py
mstone121: The ViewSet approach might not be the best approach given some of the changes suggested in https://github.com/azavea/iow-boundary-tool/pull/115. The custom actions seem flexible, but I'm guessing not flexible enough.

I think we'll be better off sticking with Simple Views.

src/django/api/views/boundary.py
```diff
@@ -0,0 +1,69 @@
...
+class BoundaryDetailView(APIView):
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request, id, format=None):
+        boundary = get_boundary_queryset_for_user(request.user)
+        boundary = boundary.select_related('utility')
+        boundary = boundary.prefetch_related(
+            Prefetch(
+                'submissions',
+                queryset=Submission.objects.select_related('review').prefetch_related(
```
jacobtylerwalls, line 59: On Slack you asked if we could make Review -> Submission one-to-one. If we make that change, that will change this clause a little bit.

src/django/api/views/boundary.py
```diff
@@ -0,0 +1,69 @@
...
+                    'review__annotations'
+                ),
+            )
+        )
+
+        boundary = boundary.filter(pk=id).first()
+        if boundary is None:
+            raise Http404
```
jacobtylerwalls, line 67: I think this would work? `boundary` is morphing from a queryset into an object, so you might consider a new name.
```suggestion
        boundary = get_object_or_404(boundary, pk=id)
```

src/django/api/management/commands/resetdb.py
```diff
@@ -22,9 +23,21 @@ def handle(self, *args, **options):
...
+        test_utility_3 = Utility(pwsid="001003", name="Azavea Test Utility 3")
+        test_utility_3.save()
+
+        test_utility_4 = Utility(pwsid="001004", name="Azavea Test Utility 4")
+        test_utility_4.save()
+
+        test_utility_5 = Utility(pwsid="001005", name="Azavea Test Utility 5")
+        test_utility_5.save()
 
         # Create test users.
```
jacobtylerwalls, line 42: Would you mind adding NC to the validator user? Otherwise the validator user can't access the API.

jacobtylerwalls (commented):
I know this is still WIP, but was hoping to this would still be helpful.

src/django/api/views/boundary.py
mstone121: It turns out that review to submission is one to one and this seems to be working correctly (or at least it's returning the right data. It might be still making too many queries).

My understanding is that Submissions and Annotations have to be prefetched because they are many to one, but reviews and utilities can be selected at the same time as their parents because they are one to one.

jacobtylerwalls: Oops, I was looking at the model defs on develop. Got it now. 👍🏻

src/django/api/views/boundary.py
```diff
@@ -0,0 +1,69 @@
...
+        boundaries = get_boundary_queryset_for_user(request.user)
+        boundaries = boundaries.select_related('utility')
+
+        requested_utilities = request.GET.get('utilities')
+        if requested_utilities:
+            for utility_id in requested_utilities:
+                if not request.user.has_access_to_utility(
+                    Utility.objects.get(pk=utility_id)
+                ):
+                    raise APIException('Cannot access this utility', 401)
```
jacobtylerwalls, line 43: This could merit updating the ADR, but if we just raise `PermissionDenied`, DRF will convert this into a 403, which I think is more accurate than 401 given that this is an insufficient permission issue rather than incorrect creds issue.

Sample response from what I was just working on in 97ef19bcb2bea89287d131b990c39931f88ad72a:
```
GET /api/boundaries/1/reference-images/
HTTP 403 Forbidden
Allow: GET, POST, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "detail": "You do not have permission to perform this action."
}
```

src/django/api/models/boundary.py
```diff
@@ -14,3 +24,42 @@ class Meta:
...
+            if review.reviewed_at is None:
+                return BOUNDARY_STATUS.IN_REVIEW
+
+            return BOUNDARY_STATUS.NEEDS_REVISIONS
+
+        return BOUNDARY_STATUS.SUBMITTED
+
+    @cached_property
+    def latest_submission(self):
+        return self.submissions.latest('created_at')
```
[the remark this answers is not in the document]
mstone121: Cool! b9b669c67fa5772d492da9e08b0f65eb1e9733f8

src/django/api/models/user.py
```diff
@@ -92,3 +92,19 @@ def clean(self):
...
+
+        if self.role == Roles.CONTRIBUTOR:
+            return utility in self.utilities.all()
+
+        if self.role == Roles.VALIDATOR:
+            all_utilites = []
+            for state in self.states.all():
+                all_utilites.extend(state.utilities.all())
+
+            return utility in all_utilites
```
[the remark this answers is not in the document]
mstone121: Yep, way better: f8cff685cff5f5306bf74f775f544996ab830294

src/django/api/views/boundary.py
mstone121: It does: ae3c80f78b5e885dde47dc4d58eb00496f457492

src/django/api/management/commands/resetdb.py
mstone121: 51871e6c36a58752a548b38e9ee417cd9904ab08

src/django/api/views/boundary.py
mstone121: I'm okay with that. Looping in @rajadain

src/django/api/views/boundary.py
```diff
@@ -0,0 +1,67 @@
...
+class BoundaryListView(APIView):
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request, format=None):
+        boundaries = get_boundary_queryset_for_user(request.user)
+        boundaries = boundaries.select_related('utility')
+
+        requested_utilities = request.GET.get('utilities')
+        if requested_utilities:
+            for utility_id in requested_utilities:
```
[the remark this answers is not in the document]
mstone121: I added it here:
[57f446c](https://github.com/azavea/iow-boundary-tool/pull/113/commits/57f446c6aab7a0809af6010b97f3d0acf835727c)
[9eb02a6](https://github.com/azavea/iow-boundary-tool/pull/113/commits/9eb02a6249155ed9798806055867bf8b8b558121)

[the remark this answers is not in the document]
mstone121: I think perhaps more validation could happen here. Passing in a value like `?utility=1|2` causes an error.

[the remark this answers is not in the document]
jacobtylerwalls: I'm okay with deferring that.

jacobtylerwalls (approved):
Looks great. We can also finalize the 403 vs. 401 thing later on.

rajadain:
Taking a look now

rajadain:
This needs the migrations to be updated:

```
Running migrations:
  No migrations to apply.
  Your models in app(s): 'api' have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

src/django/api/views/boundary.py
```diff
@@ -0,0 +1,84 @@
...
+            return None
+
+        utilities = []
+        for utility_id in utility_ids:
+            utility = get_object_or_404(Utility, pk=utility_id)
+
+            if not request.user.has_access_to_utility(utility):
+                raise APIException('Cannot access this utility', 401)
+
+            utilities.append(utility)
```
rajadain, line 56: Consider something like:

```python
if user.role == ADMINISTRATOR:
  utilities = Utilities.objects.filter(id__in=utility_ids)

if user.role == VALIDATOR:
  utilities = user.states.utilities.filter(id__in=utility_ids)

if user.role == CONTRIBUTOR:
  utilities = user.utilities.filter(id__in=utility_ids)
```

Which would get multiple utilities from the database in one go, instead of getting them one at a time with `get_object_or_404`.

rajadain: Also, we should return an empty list for non-matching utilities, instead of 404s and 401s.

rajadain (changes requested):
This looks great!

Minor comments on utility filtering, but overall this looks great. We should defer the Activity Log to a later card.

src/django/api/views/boundary.py
mstone121: These checks should be redundant because the original query set filters on user permissions. Not requiring an erro means we can just use the IDs from the query parameters directly.

8ce625e6efb687e7e3b135821721a71219125e5a

rajadain (approved):
+1 great job! Nice, clean implementation. Sets up a solid foundation for us to build upon. This looks good to merge!

Issue #81, opened 2022-09-28
Add file fields to Reference_Image and Submission


rachelekm:
## Overview

Following the discussion for issue #69 [here](https://github.com/azavea/iow-boundary-tool/issues/69#issuecomment-1260992314), we would like to track any reference images and shape files uploaded by a `User` when creating a `Submission`.  This should be done by uploading a file to an S3 bucket and storing the reference to it in the database. This issue is to add a field to the `Reference_Image` entity to track a reference image uploaded to S3 as well as a field for tracking a shapefile upload to `Submission`.

rajadain:
Reference implementation for S3 submission: https://azavea.slack.com/archives/C02N5S81A69/p1664478012094939

rajadain:
The reference image upload endpoint should take multiple images and support batch uploads.

PR #114, opened 2022-10-07
Create `ReferenceImage` serializers and views

### LINKED ISSUE #88: Wire up Draw Page to Save Data
## Overview

The Draw Page allows the Contributor to add the following pieces of data:

- Reference Images
  - File name
  - File data
  - Order in which it was added
  - Whether it is visible or not
  - Distortion
- Polygon
  - Shape
  - Optional uploaded shapefile
    - File name
    - File data
    - Whether it was edited after upload

Check with @gatesgodin whether this information should be persisted on every change, or only when the user clicks "Save and back" or "Review and submit".

Once clarified, implement the saving of this information to the relevant models on the back-end.

jacobtylerwalls:
## Overview

Create the views and serializers for ReferenceImage (everything but the image data itself, which will be in #81)

Closes #88

### Demo

Use Django admin to add a bunch of reference images to the active submission for the test utility, then visit:
Browsable APIs:
http://localhost:8181/api/boundary/1/reference-images/

### Notes
I moved the test setup script out of resetdb into `tests.py` where we can actually start using it. I had talked before about having a minimal set of users for manual testing versus a larger suite of users in python objects, but on second look, I don't see the harm in using the same set.

## Testing Instructions

- `./scripts/manage resetdb`
- Login in as contributor
- Visit API endpoints at localhost:8181/api/boundaries/pk/reference-images and localhost:8181/api/boundaries/pk/reference-images/pk
- Repeat as admin
- Repeat as validator, ensure permission denied
- Repeat with invalid data, ensure errors caught

## Checklist

- [ ] CI passes after rebase


src/django/api/urls.py
```diff
@@ -12,7 +13,8 @@
     path("auth/login/", Login.as_view()),
     path("auth/logout/", Logout.as_view()),
     path("auth/", include("dj_rest_auth.urls")),
-
+    path("referenceimages/", ReferenceImageList.as_view()),
+    path("referenceimages/<int:pk>/", ReferenceImageDetail.as_view()),
```
jacobtylerwalls, line 17: I kind of prefer `reference-images/`. Thoughts?

src/django/api/models/reference_image.py
```diff
@@ -9,7 +9,7 @@
 class ReferenceImage(models.Model):
     filename = models.CharField(max_length=255, blank=True)
     boundary = models.ForeignKey(Boundary, on_delete=models.PROTECT)
-    uploaded_at = models.DateTimeField()
+    uploaded_at = models.DateTimeField(auto_now_add=True)
```
jacobtylerwalls, line 12: Now that we intend to save to the backend on every change, I think adding a field for "order added" might be redundant. I'm thinking we can just use uploaded_at (or for race conditions, I suppose id is safer).

The "order added" on the frontend will be subordinate to the order on the server anyway, since multiple submissions for the same boundary can be worked on simultaneously by different users and be updating the backend contemporaneously.

src/django/api/tests.py
```diff
@@ -1,4 +1,211 @@
...
+        updated_image["uploaded_by"] = self.administrator.pk
+
+        response = self.client.put(update_url, updated_image, format="json")
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        ri = ReferenceImage.objects.get(pk=1)
+        self.assertIs(ri.is_visible, False)
+        self.assertEqual(ri.uploaded_by.pk, self.contributor.pk)
+
+    def test_contributor_cannot_upload_for_invalid_utility(self):
```
rajadain, line 197:
```suggestion
    def test_contributor_cannot_upload_for_invalid_boundary(self):
```

src/django/api/views/reference_image.py
```diff
@@ -0,0 +1,46 @@
...
+from ..serializers.reference_image import ReferenceImageSerializer
+
+
+class ReferenceImageList(generics.CreateAPIView):
+    permission_classes = [IsAuthenticated]
+    serializer_class = ReferenceImageSerializer
+
+    def get_queryset(self):
+        if self.request.user.role == Roles.VALIDATOR:
+            raise PermissionDenied("Validators cannot view image details.")
```
rajadain, line 18: Validators should be allowed to view images, as they may need to during inspection of the submission. They should not be allowed to edit or update them.

src/django/api/views/reference_image.py
```diff
@@ -0,0 +1,46 @@
...
+
+        boundary = get_object_or_404(Boundary, pk=self.kwargs["boundary"])
+        if boundary.utility not in self.request.user.utilities.all():
```
rajadain, line 21: This may be replaced with some of the helper methods added in #113

rajadain (commented):
Looking good overall, really nice job with the tests! Will take another look once #113 is in.

src/django/api/models/reference_image.py
rachelekm: Makes sense!

src/django/api/tests.py
```diff
@@ -1,4 +1,211 @@
...
+        url = reverse("upload_image", args=[2400])
+        with self.assertLogs(request_logger, "WARNING"):
+            response = self.client.post(url, self.initial_image, format="json")
+
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
+
+    # Validator actions
```
rachelekm, line 204: Nice tests 👏

rachelekm (approved):
Looks good! I imagine there will be some small changes when you retarget develop that might be worth a quick re-look, but the new tests work great and the put/post endpoints function as expected and error appropriately

rajadain:
Taking a look

src/django/api/views/reference_image.py
jacobtylerwalls: [507d9ba](https://github.com/azavea/iow-boundary-tool/pull/114/commits/507d9ba645ec12e67dca470dd66f5a55d4335d89)

src/django/api/views/reference_image.py
jacobtylerwalls: [507d9ba](https://github.com/azavea/iow-boundary-tool/pull/114/commits/507d9ba645ec12e67dca470dd66f5a55d4335d89)

jacobtylerwalls:
@rajadain Tested latest changes, good to merge in my opinion! Any requests for me?

rajadain (approved):
Thanks for taking a look! No, I'll rebase on `develop` and resolve the CHANGELOG issues and merge once ready.

PR #115, opened 2022-10-10
Outline REST API

### LINKED ISSUE #116: Design REST API
Design a REST API for the back-end, and document it in an ADR.

rajadain:
## Overview

Adds an outline of the REST API so we have common target to hit.

Closes #116 

## Testing Instructions

- [x] Read through the document, ensure it makes sense

## Checklist

- [ ] CI passes after rebase


doc/arch/adr-002-rest-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Authentication
+
+| Path            | Method | Body      | Authentication | Responses                                             | Notes                                        |
+| --------------- | ------ | --------- | -------------- | ----------------------------------------------------- | -------------------------------------------- |
+| `/auth/login/`  | GET    | -         | Allow Any      | 200 OK,<br />401 Not Authorized                       | Returns loging JSON if current session valid |
+| `/auth/login/`  | POST   | `{ ... }` | Allow Any      | 200 OK,<br />400 Bad Request,<br />401 Not Authorized | Used for starting new valid session          |
+| `/auth/logout/` | POST   | -         | Logged In User | 200 OK                                                | Used for logging out                         |
+
+### Boundaries
```
rajadain, line 15: I'm a little conflicted about the Boundary / Submission separation. In our data model, the Boundary is the top-level entity, which may have one or more Submissions through its life. But in the UI, we're calling everything Submissions, so this is getting a little confusing.

doc/arch/adr-002-rest-api.md
```diff
@@ -0,0 +1,74 @@
...
+| `/boundaries/`                             | GET    | -                        | Logged In User | 200 OK                                                     | Returns list of all boundaries the user has access to                                                              |
+| `/boundaries/?utilities=1,3,5,8`           | GET    | -                        | Logged In User | 200 OK,<br />401 Not Authorized                            | Returns list of boundaries in the given utilities. If the user does not have access to that utility, returns a 401 |
+| `/boundaries/`                             | POST   | `{ ... }`                | Contributor    | 201 Created,<br />400 Bad Request,<br />401 Not Authorized | Creates a new boundary if the payload is correct and the user is authorized                                        |
+| `/boundaries/{id}/`                        | GET    | -                        | Logged In User | 200 OK,<br />404 Not Found                                 | Returns boundary details if the users has access to it, else 404s. Includes all details about the boundary.        |
```
rajadain, line 22: When GETing, this would include:

- The shape
- List of reference layers
- Current status (Draft / Submitted / etc)
- Current Review Annotations
- Activity Log
- Created At, Created By, etc

doc/arch/adr-002-rest-api.md
```diff
@@ -0,0 +1,74 @@
...
+When we return `404 Not Found` for objects that the User does not have permissions to, we're using a pattern like this:
+
+```python
+@permission_classes((IsAuthenticated,))
+def get_boundary(request, id):
+    user = request.user
+    boundary = get_object_or_404(Boundary, id)
+    submission = boundary.get_latest_submission_for(user)
+
+    serializer = SubmissionSerializer(submission)
```
rajadain, line 54: This is also confusing, because the endpoint is called `get_boundary` but we're returning a serialized `submission`.

Perhaps there can be a `BoundaryDetailSerializer` that takes a `boundary` and does the `get_latest_submission_for(user)` within it, so we can return that here.

doc/arch/adr-002-rest-api.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+where `get_latest_submission_for` is along these lines:
+
+```python
+class Boundary(models.Model):
+    ...
+
+    def get_latest_submission_for(self, user):
+        if user.role == CONTRIBUTOR:
```
rajadain, line 65: Would be nice to have clean comparisons like this.

Should we consider making Role an Enum instead of a Model?

rajadain (commented):
Left some comments. Discussion encouraged.

doc/arch/adr-002-rest-api.md
jacobtylerwalls: Yes, I'm willing to do a PR for that. Make Role a varchar column on User that corresponds to the python enum we already have.

doc/arch/adr-002-rest-api.md
jacobtylerwalls: I'm willing to get rid of Boundary and just treat Utilities as boundaries (== groups of submissions).

doc/arch/adr-002-rest-api.md
```diff
@@ -0,0 +1,74 @@
...
+| `/boundaries/?utilities=1,3,5,8`           | GET    | -                        | Logged In User | 200 OK,<br />401 Not Authorized                            | Returns list of boundaries in the given utilities. If the user does not have access to that utility, returns a 401 |
+| `/boundaries/`                             | POST   | `{ ... }`                | Contributor    | 201 Created,<br />400 Bad Request,<br />401 Not Authorized | Creates a new boundary if the payload is correct and the user is authorized                                        |
+| `/boundaries/{id}/`                        | GET    | -                        | Logged In User | 200 OK,<br />404 Not Found                                 | Returns boundary details if the users has access to it, else 404s. Includes all details about the boundary.        |
+| `/boundaries/{id}/shape/`                  | PUT    | `{ ... }`                | Contributor    | 200 OK,<br />404 Not Found                                 | Updates a boundary's shape                                                                                         |
```
jacobtylerwalls, line 23: Wondering if this could just be a PATCH method on boundaries/id ?

doc/arch/adr-002-rest-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Context
+
+Based on the [data model](./adr-001-data-models.md), the following REST API design is proposed to access it:
+
+### Authentication
+
+| Path            | Method | Body      | Authentication | Responses                                             | Notes                                        |
+| --------------- | ------ | --------- | -------------- | ----------------------------------------------------- | -------------------------------------------- |
+| `/auth/login/`  | GET    | -         | Allow Any      | 200 OK,<br />401 Not Authorized                       | Returns loging JSON if current session valid |
```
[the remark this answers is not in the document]
rajadain: This should return at least the Role of the user, so decisions can be made in the front-end. It may also return the user's email, which can be shown in the UI as the currently logged in user. We don't have a spot for this in the UI, but most apps do, and we may well have one at some point.

doc/arch/adr-002-rest-api.md
```diff
@@ -0,0 +1,74 @@
...
+| `/auth/login/`  | POST   | `{ ... }` | Allow Any      | 200 OK,<br />400 Bad Request,<br />401 Not Authorized | Used for starting new valid session          |
+| `/auth/logout/` | POST   | -         | Logged In User | 200 OK                                                | Used for logging out                         |
```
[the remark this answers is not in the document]
rajadain: Good point. I think that's overkill, but I'll mention them in passing

doc/arch/adr-002-rest-api.md
rajadain: Good catch! Yes I mean reference images.

doc/arch/adr-002-rest-api.md
rajadain: It could, but PATCHes (used for partial updates) are harder to reason about than PUTs (full updates). If a Boundary had multiple editable fields, I would be more open to that, but since the shape is the only thing that can be edited in a boundary, and it may require some special handling (validating the shape, etc), I'm on the side of a dedicated endpoint.

doc/arch/adr-002-rest-api.md
```diff
@@ -0,0 +1,74 @@
...
+| `/boundaries/{id}/reference-images/`       | POST   | `{ ... }`                | Contributor    | 200 OK,<br />404 Not Found                                 | Adds a new reference image to the boundary                                                                         |
+| `/boundaries/{id}/submit/`                 | POST   | `{ notes }`              | Contributor    | 200 OK,<br />404 Not Found                                 | Submits the boundary                                                                                               |
+| `/boundaries/{id}/review/`                 | POST   | `{ notes, annotations }` | Validator      | 200 OK,<br />404 Not Found                                 | Reviews a boundary                                                                                                 |
+| `/boundaries/{id}/review/annotations/{id}` | PUT    | `{ annotation }`         | Validator      | 200 OK,<br />404 Not Found                                 | Updated an annotation                                                                                              |
```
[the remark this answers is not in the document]
rajadain: This is one of those things that I'm a little conflicted about. Given that there's multiple Reviews in the database, we _should_ have it in the endpoint to be explicit.

However, we do not allow editing old Reviews, only the current ones. Since there is only ever one active Review, we can figure it out in the back-end, and make it transparent to the front-end.

Does that make sense?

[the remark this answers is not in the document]
jacobtylerwalls: That was my understanding also. Worth a quick note in this doc, or no?

doc/arch/adr-002-rest-api.md
jacobtylerwalls: To flesh out that thought a little: otherwise, if we leave Boundary as a model, then I'm okay with the design in this ADR as is.

jacobtylerwalls (approved):

doc/arch/adr-002-rest-api.md
rajadain: A utility can have multiple boundaries submitted over the years, and each boundary is a group of submissions, so we can't make that direct link. Need something in between.

mstone121: I'm a bit confused by this. 

With `/boundaries/{id}/submit/`, is the `id` the submission ID or the boundary ID? If not the submission ID, then where is that data? In the body of the request? I don't think we can determine the submission ID just from the boundary ID.

> But in the UI, we're calling everything Submissions, so this is getting a little confusing.

The users aren't going to see the API routes, so why not use naming that reflects the data model?

mstone121: Just saw this comment: https://github.com/azavea/iow-boundary-tool/pull/115#discussion_r991251211

Is that the case with submissions too? We make them transparent to the front-end because the correct one can be determined in the back-end?

rajadain: > We make them transparent to the front-end because the correct one can be determined in the back-end?

Correct.

I'm imagining that POSTing to `/boundaries/` creates a Boundary and a Submission, and returns details for both. POSTING to `/boundaries/{id}/submit/` updates the Submission with `notes` and sets its `submitted_at` and `submitted_by`.

This reveals that we need an additional endpoint for creating a second draft submission. I'll add one.

jacobtylerwalls: But only one draft can be in review at a time? (Because 0-1 reviews per boundary.)

rajadain: Correct, only one active submission (in draft or submitted state) at a time.

mstone121 (approved):

doc/arch/adr-002-rest-api.md
jacobtylerwalls: Update from scrum: not merely one draft in review at a time, but one draft at a time _period_. Coworkers will be collaborating on the same draft.

doc/arch/adr-002-rest-api.md
```diff
@@ -0,0 +1,77 @@
...
+| `/boundaries/`                             | POST   | `{ ... }`                | Contributor    | 201 Created,<br />400 Bad Request,<br />401 Not Authorized | Creates a new boundary if the payload is correct and the user is authorized                                        |
+| `/boundaries/{id}/`                        | GET    | -                        | Logged In User | 200 OK,<br />404 Not Found                                 | Returns boundary details if the user has access to it, else 404s. Includes all details about the boundary.         |
+| `/boundaries/{id}/shape/`                  | PUT    | `{ ... }`                | Contributor    | 200 OK,<br />404 Not Found                                 | Updates the latest submission's shape                                                                              |
+| `/boundaries/{id}/reference-images/`       | POST   | `{ ... }`                | Contributor    | 200 OK,<br />404 Not Found                                 | Adds a new reference image to the boundary                                                                         |
```
jacobtylerwalls, line 26: I think we need another endpoint for updating specific reference images (visibility, distortion).

rajadain: [4e79c16](https://github.com/azavea/iow-boundary-tool/pull/115/commits/4e79c1642faea9e0bda5fe6ad924e227850b67e7)

doc/arch/adr-002-rest-api.md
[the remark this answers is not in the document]
rajadain: [d4675f2](https://github.com/azavea/iow-boundary-tool/pull/115/commits/d4675f268bbb03ed74d80efd7e4be8e457daf2d7)

doc/arch/adr-002-rest-api.md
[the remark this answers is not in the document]
rajadain: [8507d2c](https://github.com/azavea/iow-boundary-tool/pull/115/commits/8507d2c1c3475bbc7c33527417387549bdfbbf24)

rachelekm (approved):
I don't have any other comments that haven't already been discussed and clarified—thank you for writing this up!

rajadain:
Thanks for all the reviews!

Eventually we'll have a Swagger API for this, but until then we should try to keep this ADR up to date whenever we add new endpoints.

PR #117, opened 2022-10-10
Drop Role table

jacobtylerwalls:
Refactor the role field on User to be a char field having text choices, rather than pointing to a Role table.

## Overview
Follow-up to https://github.com/azavea/iow-boundary-tool/pull/115#discussion_r990863042.

## Testing Instructions

scripts/resetdb
No functional changes. Verify logging in works, Django admin choices on User form are expected.

## Checklist

- [x] CI passes after rebase


src/django/api/models/user.py
```diff
@@ -45,8 +44,13 @@ def create_superuser(self, email, password=None, **extra_fields):
...
 
-        admin_role = Role.objects.get(pk=Roles.ADMINISTRATOR)
-        return self._create_user(email, admin_role, password, **extra_fields)
+        return self._create_user(email, Roles.ADMINISTRATOR, password, **extra_fields)
+
+
+class Roles(models.TextChoices):
+    CONTRIBUTOR = "C"
+    VALIDATOR = "V"
+    ADMINISTRATOR = "A"
```
jacobtylerwalls, line 53: The example in [Django's docs](https://docs.djangoproject.com/en/3.2/ref/models/fields/#enumeration-types) uses nested classes, but I don't have a strong opinion about doing it either way.

jacobtylerwalls: Or, we could use numeric choices, but then someone reporting on the database needs to consult this repo's source to decode it.

jacobtylerwalls:
Ah,  I need to update the mermaid diagram.

Issue #96, opened 2022-09-29
Alternative Modes for Draw Page


rajadain:
## Overview

Currently, the Draw Page is used by Contributors to add Boundaries. But in the workflow, it will also be used by Contributors and Validators in two other ways:

1. **Annotation Mode**: In this mode, the user can toggle the various layers on and off, but cannot change the distortion of the reference images, nor edit the shape, nor upload a new shape. The user can add, delete, and resolve Annotations.
2. **Read Only Mode**: In this mode, the user can toggle the various layers on and off, but cannot change the distortion of the reference images, nor edit the shape, nor upload a new shape. If there are any Annotations, they can be viewed, but not deleted or resolved.

Implement these two modes. Once implemented, wire them up to the Review Map button in the Boundary Detail view, like so:

|             | Draft     | Submitted  | Needs Revisions | Approved  |
|-------------|-----------|------------|-----------------|-----------|
| **Contributor** | Editable  | Read Only  | Read Only       | Read Only |
| **Validator**   | Read Only | Annotation | Read Only       | Read Only |

PR #139, opened 2022-10-14
Load boundary details in draw page

### LINKED ISSUE #128: Draw Page: Add Boundary ID URL
## Overview

Change the Draw page to take a Boundary ID in the URL: `/draw/{id}`.

If the ID does not exist or the user does not have access to it, render a 404 page.

If the ID belongs to a Draft submission and the logged in user is a Contributor, render a regular Draw page, initialized with any information of that boundary.

If the ID belongs to a Submitted or In Review submission and the logged in user is a Validator, render an annotation only Draw page, as described in #96.

In other cases, render a read-only Draw page.

jacobtylerwalls:
## Overview

Fetch boundary details, and display them in the draw page under a new route `/draw/:id`.
Closes #128 

### Demo
Screen Shot 2022-10-14 at 12 31 06 PM

### Notes
I had to comment out `providesTags` in the relevant RTK query hook, so I'm hoping @mstone121 could point me in the right direction there.

## Testing Instructions

- `server`
- [x] Ensure 404 page for navigating to `/draw` or `/draw/99999`
- [x] Ensure welcome wizard now directs to `/draw/3` (I chose 3, because 1 was an irregular shape)
- [x] Add map button goes to welcome wizard
- [x] Boundary details are loaded on the draw tool (title, coordinates)
- [ ] Offer a view on the logic for checking roles, but most of that work is #96.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/App.js
+++ b/src/app/src/App.js
@@ -1,4 +1,3 @@
-import { Box, Flex } from '@chakra-ui/react';
 import {
     BrowserRouter,
     Outlet,
@@ -14,7 +13,7 @@ import ResetPassword from './pages/ResetPassword';
 import Welcome from './pages/Welcome';
 import Draw from './pages/Draw';
 import Submissions from './pages/Submissions';
-import Sidebar from './components/Sidebar';
+import NotFound from './pages/NotFound';
 import NavBar from './components/NavBar';
 
 import PrivateRoute from './components/PrivateRoute';
@@ -25,22 +24,14 @@ const privateRoutes = (
             <Route path='/welcome' element={<Outlet />} />
             <Route path='*' element={<NavBar />} />
         </Routes>
-        <Flex>
-            <Routes>
-                <Route path='/draw' element={<Sidebar />} />
-            </Routes>
-            <Box flex={1} position='relative'>
-                <Routes>
-                    <Route path='/welcome' element={<Welcome />} />
-                    <Route path='/draw' element={<Draw />} />
-                    <Route path='/submissions/*' element={<Submissions />} />
-                    <Route
-                        path='*'
-                        element={<Navigate to='/welcome' replace />}
-                    />
-                </Routes>
-            </Box>
-        </Flex>
+
+        <Routes>
+            <Route path='/welcome' element={<Welcome />} />
+            <Route path='/draw' element={<NotFound />} />
+            <Route path='/draw/:id' element={<Draw />} />
+            <Route path='/submissions/*' element={<Submissions />} />
+            <Route path='*' element={<Navigate to='/welcome' replace />} />
+        </Routes>
     </PrivateRoute>
 );
 
--- a/src/app/src/components/DrawTools/DrawTools.js
+++ b/src/app/src/components/DrawTools/DrawTools.js
@@ -1,3 +1,6 @@
+import { useEffect } from 'react';
+import { useDispatch } from 'react-redux';
+
 import { Button, Icon } from '@chakra-ui/react';
 import { ArrowRightIcon } from '@heroicons/react/outline';
 
@@ -9,7 +12,28 @@ import useEditingPolygon from './useEditingPolygon';
 import useGeocoderResult from './useGeocoderResult';
 import useTrackMapZoom from './useTrackMapZoom';
 
-export default function DrawTools() {
+import { setPolygon } from '../../store/mapSlice';
+
+export default function DrawTools({ details }) {
+    const dispatch = useDispatch();
+
+    // Add the polygon indicated by `details` to the state
+    useEffect(() => {
+        if (details) {
+            dispatch(
+                setPolygon({
+                    // endpoint returns lngLat, leaflet needs latLng
+                    points: details.submission.shape.coordinates[0].map(p => [
+                        p[1],
+                        p[0],
+                    ]),
+                    visible: true,
+                    label: details.utility.name,
+                })
+            );
+        }
+    }, [dispatch, details]);
+
     useEditingPolygon();
     useAddPolygonCursor();
     useGeocoderResult();
--- a/src/app/src/constants.js
+++ b/src/app/src/constants.js
@@ -1,4 +1,4 @@
-export const MAP_CENTER = [35.7596, -79.0193]; // North Carolina
+export const MAP_CENTER = [35.7938, -78.7232]; // Raleigh, North Carolina
 export const MAP_INITIAL_ZOOM = 13;
 export const INITIAL_POLYGON_SCALE_FACTOR = 0.5;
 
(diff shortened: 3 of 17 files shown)


src/app/src/api/boundaries.js
```diff
@@ -21,7 +21,7 @@ const boundaryApi = api.injectEndpoints({
                 url: `/boundaries/${id}/`,
                 method: 'GET',
             }),
-            providesTags: getSingleItemProvider(TAGS.BOUNDARY),
+            // providesTags: getSingleItemProvider(TAGS.BOUNDARY),
```
jacobtylerwalls, line 24: @mstone121 I would be grateful if you have pointers for debugging the failure I saw:

`Uncaught (in promise) TypeError: description(...).map is not a function`

src/app/src/App.js
```diff
@@ -25,22 +24,14 @@ const privateRoutes = (
             <Route path='/welcome' element={<Outlet />} />
             <Route path='*' element={<NavBar />} />
         </Routes>
-        <Flex>
-            <Routes>
-                <Route path='/draw' element={<Sidebar />} />
-            </Routes>
```
jacobtylerwalls, line 31: This refactor was to avoid rendering a sidebar for the 404 not found.

src/app/src/constants.js
```diff
@@ -1,4 +1,4 @@
-export const MAP_CENTER = [35.7596, -79.0193]; // North Carolina
+export const MAP_CENTER = [35.7938, -78.7232]; // Raleigh, North Carolina
```
jacobtylerwalls, line 1: This slight nudge was so that the test submission would be visible by default. Better would be to add a `flyTo` call, but I figured that might be its own card (center, zoom level etc could be relevant).

src/app/src/pages/Draw.js
```diff
@@ -1,12 +1,67 @@
...
+        ) &&
+        user.role === ROLES.VALIDATOR
+    ) {
+        mode = DRAW_MODES.ANNOTATIONS_ONLY;
+    } else if (
+        details.status === BOUNDARY_STATUS.DRAFT &&
+        user.role === ROLES.CONTRIBUTOR
+    ) {
+        mode = DRAW_MODES.FULLY_EDITABLE;
+    }
```
jacobtylerwalls, line 54: This might fit better in `<DrawTools>` or else just leave it entirely to the implementer of #96.

src/app/src/api/boundaries.js
mstone121: This is from my implementation. I forgot that when `providesTags` or `invalidatesTags` is passed a function, it expects that function to return an array of tags rather than a single tags. This diff should fix the issue:

```diff
diff --git a/src/app/src/api/boundaries.js b/src/app/src/api/boundaries.js
index 0199429..c362afd 100644
--- a/src/app/src/api/boundaries.js
+++ b/src/app/src/api/boundaries.js
@@ -21,7 +21,7 @@ const boundaryApi = api.injectEndpoints({
                 url: `/boundaries/${id}/`,
                 method: 'GET',
             }),
-            // providesTags: getSingleItemProvider(TAGS.BOUNDARY),
+            providesTags: getSingleItemProvider(TAGS.BOUNDARY),
         }),
 
         startNewBoundary: build.mutation({
diff --git a/src/app/src/api/tags.js b/src/app/src/api/tags.js
index 9903447..5f3146c 100644
--- a/src/app/src/api/tags.js
+++ b/src/app/src/api/tags.js
@@ -12,15 +12,15 @@ export function getListTagProvider(tag) {
 }
 
 export function getSingleItemProvider(tag) {
-    return (result, error, id) => ({ tag, id });
+    return (result, error, id) => [{ tag, id }];
 }
 
 export function getNewItemTagInvalidator(tag) {
-    return () => ({ tag, id: TAGS.LIST_ID });
+    return () => [{ tag, id: TAGS.LIST_ID }];
 }
 
 export function getUpdateItemTagInvalidator(tag) {
-    return (result, error, id) => ({ tag, id });
+    return (result, error, id) => [{ tag, id }];
 }
 
 export default TAGS;

File: src/app/src/components/DrawTools/DrawTools.js [hunk 5]
```diff
@@ -9,7 +12,27 @@ import useEditingPolygon from './useEditingPolygon';
...
-export default function DrawTools() {
+import { addPolygon } from '../../store/mapSlice';
+
+export default function DrawTools({ details }) {
+    const dispatch = useDispatch();
+
+    useEffect(() => {
+        if (details) {
+            dispatch(
+                addPolygon({
```
mstone121, line 23: Unrelated to your changes, but at first I was worried this would add multiple polygons (since details is an object and won't memoize accross renders), but it turns out `addPolygon` is overriding the polygon in state anyway. Maybe this would be better if named `setPolygon`?

src/app/src/constants.js
```diff
@@ -65,3 +65,10 @@ export const BOUNDARY_STATUS = {
     NEEDS_REVISIONS: 'Needs Revisions',
     APPROVED: 'Approved',
 };
+
+// In sync with django.api.models.user.Roles
```
mstone121, line 69: Can we add a similar comment to the django enum?

mstone121 (approved):
Looks good and working well. I commented with a diff to fix the tag problem.

jacobtylerwalls:
Thanks @mstone121, very helpful--I pushed up edits.

src/app/src/components/DrawTools/DrawTools.js
```diff
@@ -9,7 +12,27 @@ import useEditingPolygon from './useEditingPolygon';
 import useGeocoderResult from './useGeocoderResult';
 import useTrackMapZoom from './useTrackMapZoom';
 
-export default function DrawTools() {
+import { setPolygon } from '../../store/mapSlice';
+
+export default function DrawTools({ details }) {
+    const dispatch = useDispatch();
+
+    useEffect(() => {
```
rajadain, line 20: Consider adding a comment block here summarizing what this `useEffect` does, which will be handy for future reviews.

src/app/src/pages/Draw.js
```diff
@@ -1,12 +1,67 @@
...
+        return (
+            <Box w='100%' h='100vh'>
+                <Flex direction='column' alignItems='center'>
+                    <Spinner mt={60} />
+                </Flex>
+            </Box>
+        );
+    }
+
+    if (error) {
```
rajadain, line 36:
```suggestion
    if (error || typeof details !== 'object') {
```

This will prevent errors from pages like http://localhost:4545/draw/xyz, which send back an HTML response instead of JSON. This is because a lot of 404s are handled by [Django SPA Middleware](https://django-spa.readthedocs.io/en/latest/readme.html) that sends all unknown URLs to the front-end to be resolved by the client router.

rajadain (commented):
This is looking good! Left a couple minor comments.

jacobtylerwalls:
Thanks for the reviews!

PR #140, opened 2022-10-14
Add Activity Log Serializer

### LINKED ISSUE #125: Add Serializer for Activity Log
## Overview

Add a serializer for the Activity Log, and include its data in the `/boundaries/{id}/` endpoint.

It should be a list of activities, ordered most recent first. Each activity should have an action ("Approved" "Reviewed" "Submitted" "Drafted"), an actor (user's name), and a date. Some activities may also have notes. Reviews will also have annotations.

rachelekm:
## Overview

Adds activity log data as part of the `/boundaries/{id}/` endpoint. Each activity in the log should include the activity type, user taking the action, and time of action as well as any notes and, if an activity is submitting a review, number of annotations. Activity types should be “Drafted", "Submitted", "Reviewed", "Approved".

Closes #125 

### Demo

Screen Shot 2022-10-14 at 6 42 45 PM

## Testing Instructions

I updated the test data to include a handful more submissions to better test the ability to order log events. The below endpoint includes multiple submissions:
http://localhost:8181/api/boundaries/4/

- `scripts/resetdb`
- Login and visit the above api endpoints
- Confirm the activity log is returned as expected
- Login to admin and create a couple various submission activities for a boundary
- Confirm activity log updates correctly and in order for that boundary detail endpoint
- `scripts/test` to confirm no errors

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Wire up Boundary List Page [#122](https://github.com/azavea/iow-boundary-tool/pull/122)
 - Add ReferenceImage views/serializers for image metadata [#114](https://github.com/azavea/iow-boundary-tool/pull/114)
 - Load boundary details in draw page [#139](https://github.com/azavea/iow-boundary-tool/pull/139)
+- Add Activity Log Serializer [#140](https://github.com/azavea/iow-boundary-tool/pull/140)
 
 ### Changed
 
--- a/src/django/api/serializers/activity_log.py
+++ b/src/django/api/serializers/activity_log.py
@@ -1,0 +1,70 @@
+from rest_framework.serializers import (
+    ModelSerializer,
+    SerializerMethodField,
+    EmailField,
+    DateTimeField,
+)
+
+from ..models import Submission
+
+
+class ActivityDraftedSerializer(ModelSerializer):
+    user = EmailField(source='created_by.email')
+    action = SerializerMethodField()
+    time = DateTimeField(source='created_at')
+    notes = ''
+
+    def get_action(self, value):
+        return 'Drafted'
+
+    class Meta:
+        model = Submission
+        fields = ['user', 'action', 'time']
+        optional_fields = ['annotations_count', 'notes']
+
+
+class ActivitySubmittedSerializer(ActivityDraftedSerializer):
+    def to_representation(self, instance):
+        obj = super().to_representation(instance)
+        obj['action'] = 'Submitted'
+        obj['time'] = instance.submitted_at
+        obj['user'] = instance.submitted_by.email
+        obj['notes'] = instance.notes
+        return obj
+
+
+class ActivityReviewStartedSerializer(ActivityDraftedSerializer):
+    def to_representation(self, instance):
+        obj = super().to_representation(instance)
+        review = instance.review
+        obj['action'] = 'Review Started'
+        obj['time'] = review.created_at
+        obj['user'] = review.reviewed_by.email
+        # Don't show review notes in log until review is complete
+        if hasattr(obj, 'notes'):
+            obj.pop('notes')
+        return obj
+
+
+class ActivityReviewedSerializer(ActivityReviewStartedSerializer):
+    def to_representation(self, instance):
+        obj = super().to_representation(instance)
+        review = instance.review
+        obj['action'] = 'Reviewed'
+        obj['time'] = review.reviewed_at
+        obj['notes'] = review.notes
+        obj['annotations_count'] = review.annotations.count()
+        return obj
+
+
+class ActivityApprovedSerializer(ActivityDraftedSerializer):
+    def to_representation(self, instance):
+        obj = super().to_representation(instance)
+        approval = instance.approval
+        if approval.approved_at:
+            obj['action'] = 'Approved'
+            obj['time'] = approval.approved_at
+            obj['user'] = approval.approved_by.email
+            if hasattr(obj, 'notes'):
+                obj.pop('notes')
+        return obj
--- a/src/django/api/tests.py
+++ b/src/django/api/tests.py
@@ -61,90 +61,130 @@ class BoundarySyncAPITestCase(TestCase):
         cls.boundary_4 = Boundary.objects.create(utility=test_utility)
         cls.boundary_5 = Boundary.objects.create(utility=test_utility)
 
+        # Submissions activities should follow
+        # Draft-->Submission-->Approved/Revisions flow
+        # Order of test data activities dependent on order of below methods executing,
+        # So there is potential for order variation if unexpected execution sequence.
+        # If variation, explicitly override the created_at value on Submission creation
+
         # draft
         Submission.objects.create(
             boundary=cls.boundary_1,
             shape=RALEIGH_FAKE_ZIGZAG,
-            created_at=datetime(2022, 10, 3, hour=15, tzinfo=timezone.utc),
+            created_by=cls.contributor,
+        )
+
+        submission_2 = Submission.objects.create(
+            boundary=cls.boundary_2,
+            shape=RALEIGH_FAKE_TRIANGLE,
+            created_by=cls.contributor,
+        )
+
+        submission_3 = Submission.objects.create(
+            boundary=cls.boundary_3,
+            shape=RALEIGH_FAKE_TRIANGLE,
+            created_by=cls.contributor,
+        )
+
+        submission_4 = Submission.objects.create(
+            boundary=cls.boundary_4,
+            shape=RALEIGH_FAKE_TRIANGLE,
+            created_by=cls.contributor,
+        )
+
+        submission_5 = Submission.objects.create(
+            boundary=cls.boundary_5,
+            shape=RALEIGH_FAKE_RECTANGLE,
             created_by=cls.contributor,
         )
 
         # submitted
-        Submission.objects.create(
-            boundary=cls.boundary_2,
-            shape=RALEIGH_FAKE_TRIANGLE,
-            created_at=datetime(2022, 10, 2, hour=11, tzinfo=timezone.utc),
-            created_by=cls.contributor,
-            submitted_at=datetime(2022, 10, 2, hour=12, tzinfo=timezone.utc),
+        Submission.objects.filter(pk=submission_2.pk).update(
+            submitted_at=datetime.now(tz=timezone.utc),
+            submitted_by=cls.contributor,
+            notes="Notes for the test submission.",
+        )
+
+        Submission.objects.filter(pk=submission_3.pk).update(
+            submitted_at=datetime.now(tz=timezone.utc),
+            submitted_by=cls.contributor,
+            notes="Notes for the test submission.",
+        )
+
+        Submission.objects.filter(pk=submission_4.pk).update(
+            submitted_at=datetime.now(tz=timezone.utc),
+            submitted_by=cls.contributor,
+            notes="Notes for the test submission.",
+        )
+
+        Submission.objects.filter(pk=submission_5.pk).update(
+            submitted_at=datetime.now(tz=timezone.utc),
             submitted_by=cls.contributor,
             notes="Notes for the test submission.",
         )
 
         # reviewing
-        submitted = Submission.objects.create(
-            boundary=cls.boundary_3,
-            shape=RALEIGH_FAKE_TRIANGLE,
-            created_at=datetime(2022, 10, 2, hour=4, tzinfo=timezone.utc),
-            created_by=cls.contributor,
-            submitted_at=datetime(2022, 10, 2, hour=6, tzinfo=timezone.utc),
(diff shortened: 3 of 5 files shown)


src/django/api/tests.py
```diff
@@ -69,6 +69,17 @@ def setUpTestData(cls):
...
+        draft = Submission.objects.create(
+            boundary=cls.boundary_2,
+            shape=RALEIGH_FAKE_ZIGZAG,
+            created_by=cls.contributor,
+        )
+
+        # Created_at auto-populates so need to update to earlier date
+        Submission.objects.filter(pk=draft.pk).update(
+            created_at=datetime(2022, 10, 2, hour=9, tzinfo=timezone.utc)
+        )
```
jacobtylerwalls, line 81: At your option, as `update` feels more of a bulk-operation thing.
```suggestion
        draft.created_at = datetime(2022, 10, 2, hour=9, tzinfo=timezone.utc)
        draft.save()
```

src/django/api/tests.py
```diff
@@ -131,6 +142,17 @@ def setUpTestData(cls):
             created_at=datetime(2022, 10, 2, hour=10, minute=5, tzinfo=timezone.utc),
```
jacobtylerwalls, line 142: Not related to your changes, but found in testing. Would you mind adding `reviewed_at` to the the review object, a few lines above? We may not need both `created_at` and `reviewed_at`, but that's another discussion.

src/django/api/tests.py
```diff
@@ -148,6 +170,22 @@ def setUpTestData(cls):
             approved_by=cls.validator,
         )
 
+        approved = Submission.objects.create(
```
jacobtylerwalls, line 173: IMO we shouldn't overwrite the `approved` variable name, which can get confusing! (I see this was already happening with some names here.)

src/django/api/serializers/activity_log.py
```diff
@@ -0,0 +1,48 @@
...
+                obj['notes'] = review.notes
+                obj['annotations_count'] = review.annotations.count()
+        if hasattr(instance, 'approval'):
+            approval = instance.approval
+            approver = User.objects.get(pk=approval.approved_by.pk)
+            if approval.approved_at:
+                obj['action'] = 'Approved'
+                obj['time'] = approval.approved_at
+                obj['user'] = approver.email
+        return obj
```
jacobtylerwalls, line 48: I noticed reviews and approvals are blocking activity entries for submissions/drafts. For instance, if you visit http://localhost:8181/api/boundaries/5/, you get a single entry for an approval. I would have expected to have also seen the drafted, submitted, etc.

We can exchange some solutions over Slack?

src/django/api/views/boundary.py
```diff
@@ -54,8 +54,16 @@ def get(self, request, id, format=None):
         boundary_set = boundary_set.prefetch_related(
             Prefetch(
                 'submissions',
-                queryset=Submission.objects.select_related('review').prefetch_related(
-                    'review__annotations'
+                queryset=Submission.objects.select_related('review')
+                .select_related('approval')
```
jacobtylerwalls, line 58: Purely cosmetic AFAICT, but you can just `select_related('review', 'approval')`

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Configure Amazon SES email support in staging env [#118](https://github.com/azavea/iow-boundary-tool/pull/118)
 - Wire up Boundary List Page [#122](https://github.com/azavea/iow-boundary-tool/pull/122)
 - Add ReferenceImage views/serializers for image metadata [#114](https://github.com/azavea/iow-boundary-tool/pull/114)
+- Add Add Activity Log Serializer [#140](https://github.com/azavea/iow-boundary-tool/pull/140)
```
jacobtylerwalls, line 45:
```suggestion
- Add Activity Log Serializer [#140](https://github.com/azavea/iow-boundary-tool/pull/140)
```

jacobtylerwalls (changes requested):
I just had one question about capturing the original drafted/submitted actions after a review or approval of a boundary. The rest is cosmetic feedback at your option.

src/django/api/tests.py
jacobtylerwalls: At some future point we might create a datetime helper that could take just a number of hours after 10/1/2022 and produce the wanted datetimes.

rajadain: We're also specifying a value for `created_at` in the creation of all other instances here. Do they also not work correctly? Or can we just use do that?

rajadain: Perhaps something like this:

```python
def dt(date: str, hour: int = 0) -> datetime:
    """
    Turns a string of type "YYYY-MM-DD" into a datetime.
    Uses the optional hour value if specified, else defaults to 0.
    """
    year = int(date[:4])
    month = int(date[5:7])
    day = int(date[8:])

    return datetime(year, month, day, hour, tzinfo=timezone.utc)
```
```python
   created_at=dt('2022-10-03', hour=15),
```

rajadain (commented):
Same here, that I'm not getting the full activity log for the approved `boundaries/5`. Should see when it was first drafted, submitted, reviewed, drafted again, submitted again, and approved.

src/django/api/tests.py
jacobtylerwalls: > We're also specifying a value for created_at in the creation of all other instances here. Do they also not work correctly? Or can we just use do that?

I did it that way to reduce the variation between test runs. If it became important to structure test cases so that action B took place in between actions A and C, I didn't want to rely on the microsecond timing between parts of this test method executing.

I might be missing your point, though!

rajadain: > I might be missing your point, though!

My point is specifying `created_at` in

```python
Submission.objects.create(created_at=...)
```

vs changing the value after creation:

```python
s = Submission.objects.create(...)
s.created_at = ...
s.save()
```

Does that make sense?

jacobtylerwalls: Ah, yes it does!

rajadain: So should we change it everywhere then? Not just here?

rajadain: Alternatively, if we just created the items in the right order, that should be enough, and we may drop `created_at=` specification completely.

rachelekm: Ah, good call. The latter option seems like it would be more intuitive/readable, so I'm for that! However I understand Jacob's point on reducing potential variations. Maybe a compromise is dropping the specification to improve readability and adding a test to check that the order of execution of submissions creation is correct?

jacobtylerwalls: I retract my concern :-)

For now I bet a comment is fine, we can always add the test later.

rachelekm: Thanks all, will do a comment! If there's time an activity log test could potentially be dual-purpose and confirm serializer output as well as the correct order, too.

src/django/api/tests.py
rachelekm: Whoops, good catch!

src/django/api/serializers/activity_log.py
```diff
@@ -0,0 +1,48 @@
...
+    def get_action(self, value):
+        return 'Drafted'
+
+    class Meta:
+        model = Submission
+        fields = ['user', 'action', 'time', 'notes']
+        optional_fields = ['annotations_count']
+
+    def to_representation(self, instance):
+        obj = super(ActivityLogSerializer, self).to_representation(instance)
```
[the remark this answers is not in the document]
rachelekm: oh that's helpful clarification, thank you!

src/django/api/tests.py
rachelekm: Since we are adding a "Review Started" activity log for now, that object works well as a "review started but not yet submitted" test, so I left it as is for now.

rachelekm:
Thanks, all! Edits are in and ready for another round of reviews. As discussed on slack I added in "Review Started" as another action to log—I left off any notes/annotations for those as it seemed like something only worth displaying once the review is submitted.

src/django/api/serializers/activity_log.py
```diff
@@ -0,0 +1,68 @@
...
+        obj = super().to_representation(instance)
+        review = instance.review
+        obj['action'] = 'Reviewed'
+        obj['time'] = review.reviewed_at
+        obj['notes'] = review.notes
+        obj['annotations_count'] = review.annotations.count()
+        return obj
+
+
+class ActivityApprovedSerializer(ActivityDraftedSerializer):
```
rajadain, line 59: Approvals should not have any notes:

image

src/django/api/serializers/activity_log.py
```diff
@@ -0,0 +1,68 @@
...
+        obj['time'] = instance.submitted_at
+        obj['user'] = submitter.email
+        return obj
+
+
+class ActivityReviewStartedSerializer(ActivityDraftedSerializer):
+    def to_representation(self, instance):
+        obj = super().to_representation(instance)
+        review = instance.review
+        reviewer = User.objects.get(pk=review.reviewed_by.pk)
```
rajadain, line 39: Seeing the contributor as the user for Review activities: http://localhost:8181/api/boundaries/4/

image

It should be the validator.

src/django/api/serializers/activity_log.py
```diff
@@ -0,0 +1,68 @@
...
+    user = EmailField(source='created_by.email')
+    action = SerializerMethodField()
+    time = DateTimeField(source='created_at')
+
+    def get_action(self, value):
+        return 'Drafted'
+
+    class Meta:
+        model = Submission
+        fields = ['user', 'action', 'time', 'notes']
```
rajadain, line 21: Drafts should not have notes:

image

src/django/api/tests.py
```diff
@@ -61,90 +61,130 @@ def setUpTestData(cls):
         cls.boundary_4 = Boundary.objects.create(utility=test_utility)
         cls.boundary_5 = Boundary.objects.create(utility=test_utility)
 
+        # Submissions activities should follow
+        # Draft-->Submission-->Approved/Revisions flow
+        # Order of test data activities dependent on order of below methods executing,
+        # So there is otentional for order variation if unexpected execution sequence
```
rajadain, line 67:
```suggestion
        # So there is potential for order variation if unexpected execution sequence
```

rajadain (changes requested):
Code organization looks good! Some inaccuracies in the output though.

rachelekm:
Thanks for catching those! Just made the `notes` field optional too since there's no reason to send back an empty string anymore.

src/django/api/serializers/activity_log.py
```diff
@@ -0,0 +1,73 @@
...
+        optional_fields = ['annotations_count', 'notes']
+
+
+class ActivitySubmittedSerializer(ActivityDraftedSerializer):
+    def to_representation(self, instance):
+        obj = super().to_representation(instance)
+        submitter = User.objects.get(pk=instance.submitted_by.pk)
+        obj['action'] = 'Submitted'
+        obj['time'] = instance.submitted_at
+        obj['user'] = submitter.email
```
rajadain, line 32: This can be simplified to

```suggestion
        obj['user'] = instance.submitted_by.email
```

Don't need to

```python
submitter = User.objects.get(pk=instance.submitted_by.pk)
```

src/django/api/tests.py
```diff
@@ -61,90 +61,130 @@ def setUpTestData(cls):
         cls.boundary_4 = Boundary.objects.create(utility=test_utility)
         cls.boundary_5 = Boundary.objects.create(utility=test_utility)
 
+        # Submissions activities should follow
+        # Draft-->Submission-->Approved/Revisions flow
+        # Order of test data activities dependent on order of below methods executing,
+        # So there is potentional for order variation if unexpected execution sequence
```
rajadain, line 67:
```suggestion
        # So there is potential for order variation if unexpected execution sequence
```

rajadain (approved):
+1 nice work on this! Left a couple minor code comments, thence this is good to merge.

src/django/api/tests.py
rachelekm: Only minorly embarrassed this got me twice 🤦‍♀️

jacobtylerwalls (approved):
Looks great!

src/django/api/tests.py
```diff
@@ -61,90 +61,130 @@ def setUpTestData(cls):
...
             submitted_by=cls.contributor,
             notes="Notes for the test submission.",
         )
 
-        review = Review.objects.create(
-            submission=submitted,
-            reviewed_by=cls.validator,
-            notes="Notes for the review.",
-            created_at=datetime(2022, 10, 2, hour=10, tzinfo=timezone.utc),
+        Submission.objects.filter(pk=submission_3.pk).update(
```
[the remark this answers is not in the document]
rachelekm: Good catch! That'll be a nice cleanup. I'll leave it for now as you mentioned since we'll go back over these in the next couple cards.

Issue #132, opened 2022-10-13
Draw Page: Create Submission Form


rajadain:
## Overview

When a user is ready to submit a boundary, they should see this form:

![image](https://user-images.githubusercontent.com/1430060/195501241-04f91643-d2ff-485d-a0ce-5c4a3b258c9e.png)

![image](https://user-images.githubusercontent.com/1430060/195691601-f8630517-8dbf-4c26-9307-9daad950d919.png)

where the date and utility details are prefilled.

Clicking Submit for Review should update the submission, add notes, and populate the submitted at and submitted by fields.

Issue #127, opened 2022-10-13
Add MultiPolygon Support to Submission Shapes


rajadain:
## Overview

Currently, a submission is only allowed to have simple polygon shapes:

https://github.com/azavea/iow-boundary-tool/blob/100e8d049fa95d43e05ff63744f1ef9acb23332a/src/django/api/models/submission.py#L17

However, a number of real world utility boundaries are complex multipolygons. While we may not support drawing complex multipolygons at this stage, we should still allow users to upload shapefiles or geojsons created elsewhere. To store this, we'll need to change this field to be a MultiPolygon.

Also update any test data and tests to match this.

PR #141, opened 2022-10-18
Save shape updates

### LINKED ISSUE #130: Draw Page: Update Shape When Changed
## Overview

Whenever the shape is updated or changed in the Draw Page, or a new file is uploaded to replace existing shape, update this in the database.

mstone121:
## Overview

This PR adds functionality that saves a boundary when it is updated.

Updating is debouced with a three second interval. Each update is patched immediately and directly into the RTK query cache. After the three second interval, the latest update is sent to the server. If the server fails, changes are rolled back to the last saved state.

Closes #130 

### Notes

The map is re-rendering on update. I believe this is because a component above the map is using the boundary details query. When the shape is updated, this query is updated, which triggers a re-render. This should be addressed here if possible or in a later card.

## Testing Instructions

- Log in as c1@azavea.com
- Go to http://localhost:4545/draw/1
- Make a few adjustments to the shape, wait 3 seconds, and reload the page
  - [x] Ensure updates were saved
- Repeat with v1@azavea.com and http://localhost:4545/draw/3
  - [ ] Ensure updates were saved

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/components/DrawTools/DrawTools.js
+++ b/src/app/src/components/DrawTools/DrawTools.js
@@ -1,5 +1,5 @@
 import { useEffect } from 'react';
-import { useDispatch } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 
 import { Button, Icon } from '@chakra-ui/react';
 import { ArrowRightIcon } from '@heroicons/react/outline';
@@ -7,14 +7,45 @@ import { ArrowRightIcon } from '@heroicons/react/outline';
 import EditToolbar from './EditToolbar';
 import MapControlButtons from './MapControlButtons';
 
+import { useGetBoundaryDetailsQuery } from '../../api/boundaries';
+import { useBoundaryId, useEndpointToastError } from '../../hooks';
+
 import useAddPolygonCursor from './useAddPolygonCursor';
 import useEditingPolygon from './useEditingPolygon';
 import useGeocoderResult from './useGeocoderResult';
 import useTrackMapZoom from './useTrackMapZoom';
 
 import { setPolygon } from '../../store/mapSlice';
+import { BOUNDARY_STATUS, ROLES } from '../../constants';
+import LoadingModal from '../LoadingModal';
 
-export default function DrawTools({ details }) {
+const DRAW_MODES = {
+    FULLY_EDITABLE: 'fully_editable',
+    ANNOTATIONS_ONLY: 'annotations_only',
+    READ_ONLY: 'read_only',
+};
+
+export default function LoadBoundaryDetails() {
+    const user = useSelector(state => state.auth.user);
+    const id = useBoundaryId();
+
+    const { isFetching, data: details, error } = useGetBoundaryDetailsQuery(id);
+    useEndpointToastError(error);
+
+    if (isFetching) {
+        return <LoadingModal isOpen title='Loading boundary data...' />;
+    }
+
+    if (error || typeof details !== 'object') {
+        return null;
+    }
+
+    const mode = getDrawMode({ status: details.status, userRole: user.role });
+
+    return <DrawTools mode={mode} details={details} />;
+}
+
+function DrawTools({ mode, details }) {
     const dispatch = useDispatch();
 
     // Add the polygon indicated by `details` to the state
@@ -43,6 +74,18 @@ export default function DrawTools({ details }) {
             <MapControlButtons />
         </>
     );
+}
+
+function getDrawMode({ status, userRole }) {
+    if (userRole === ROLES.VALIDATOR && status === BOUNDARY_STATUS.IN_REVIEW) {
+        return DRAW_MODES.ANNOTATIONS_ONLY;
+    }
+
+    if (status === BOUNDARY_STATUS.DRAFT && userRole === ROLES.CONTRIBUTOR) {
+        return DRAW_MODES.FULLY_EDITABLE;
+    }
+
+    return DRAW_MODES.READ_ONLY;
 }
 
 function SaveAndBackButton() {
--- a/src/app/src/components/DrawTools/DrawTools.js
+++ b/src/app/src/components/DrawTools/DrawTools.js
@@ -22,11 +22,7 @@ export default function DrawTools({ details }) {
         if (details) {
             dispatch(
                 setPolygon({
-                    // endpoint returns lngLat, leaflet needs latLng
-                    points: details.submission.shape.coordinates[0].map(p => [
-                        p[1],
-                        p[0],
-                    ]),
+                    points: details.submission.shape.coordinates[0],
                     visible: true,
                     label: details.utility.name,
                 })
--- a/src/app/src/hooks.js
+++ b/src/app/src/hooks.js
@@ -9,7 +9,6 @@ import {
     updateReferenceImage,
 } from './store/mapSlice';
 import { useParams } from 'react-router';
-import api from './api/api';
 
 export function useDialogController() {
     const [isOpen, setIsOpen] = useState(false);
@@ -151,29 +150,6 @@ export function useTrailingDebounceCallback({
         },
         [callback, immediateCallback, interval]
     );
-}
-
-export function useReverseQueue({ queryArgs, endpointName }) {
-    const dispatch = useDispatch();
-
-    const queue = useRef([]);
-
-    const clear = useCallback(() => {
-        queue.current = [];
-    }, []);
-
-    const apply = useCallback(() => {
-        dispatch(
-            api.util.patchQueryData(endpointName, queryArgs, queue.current)
-        );
-        clear();
-    }, [dispatch, queryArgs, endpointName, clear]);
-
-    const push = useCallback(patches => {
-        queue.current = [...patches, ...queue.current];
-    }, []);
-
-    return { queue, clear, apply, push };
 }
 
 export function useEndpointToastError(error, message = 'An error occured') {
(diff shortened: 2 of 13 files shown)


src/app/src/hooks.js
```diff
@@ -110,3 +112,54 @@ export function useFilePicker(onChange) {
 
     return openFileDialog;
 }
+
+export function useBoundaryId() {
+    return useParams().boundaryId;
+}
+
+export function useTrailingDebouceCallback({
```
jacobtylerwalls, line 120:
```suggestion
export function useTrailingDebounceCallback({
```

src/django/api/views/boundary.py
```diff
@@ -63,3 +68,50 @@ def get(self, request, id, format=None):
...
+
+    def post(self, request, id, format=None):
+        boundary_set = get_boundary_queryset_for_user(request.user)
+        boundary_set = boundary_set.prefetch_related('submissions')
+        boundary = get_object_or_404(boundary_set, pk=id)
+
+        if boundary.status == BOUNDARY_STATUS.DRAFT:
+            if request.user.role not in [Roles.CONTRIBUTOR, Roles.ADMINISTRATOR]:
+                raise ForbiddenException(
+                    'Only contributors and administrators can edit only drafts.'
```
jacobtylerwalls, line 84:
```suggestion
                    'Only contributors and administrators can edit drafts.'
```

src/django/api/views/boundary.py
```diff
@@ -63,3 +68,50 @@ def get(self, request, id, format=None):
...
+                    'Only validators and administrators can edit boundaries in review.'
+                )
+
+            shape = BoundaryShapeView.get_valid_shape(request.data)
+
+            boundary.latest_submission.review.shape = shape
+            boundary.latest_submission.review.save()
+
+        else:
+            return BadRequestException(
```
rajadain, line 104:
```suggestion
            raise BadRequestException(
```

src/django/api/serializers/shape.py
```diff
@@ -0,0 +1,32 @@
...
+    )
+
+    def to_internal_value(self, data):
+        validated_data = super().to_internal_value(data)
+        coordinates = validated_data['coordinates']
+
+        if not ShapeSerializer.coordinates_are_closed(coordinates):
+            coordinates = ShapeSerializer.get_closed_coordinates(coordinates)
+
+        return Polygon(*coordinates)
```
rajadain, line 23: This may be opposed to / need to be overridden when #127 is implemented.

src/django/api/exceptions.py
```diff
@@ -0,0 +1,13 @@
...
+class ForbiddenException(APIException):
+    status_code = 403
+    default_detail = 'You are not allowed to perform this action.'
+    default_code = 'forbidden'
+
+
+class BadRequestException(APIException):
+    status_code = 400
+    default_detail = 'There was a problem with your request.'
+    default_code = 'bad_request'
```
rajadain, line 13: Nice job adding clear exceptions

src/django/api/views/boundary.py
```diff
@@ -63,3 +68,50 @@ def get(self, request, id, format=None):
...
+
+            shape = BoundaryShapeView.get_valid_shape(request.data)
+
+            boundary.latest_submission.shape = shape
+            boundary.latest_submission.save()
+
+        elif boundary.status == BOUNDARY_STATUS.IN_REVIEW:
+            if request.user.role not in [Roles.VALIDATOR, Roles.ADMINISTRATOR]:
+                raise ForbiddenException(
+                    'Only validators and administrators can edit boundaries in review.'
```
rajadain, line 95: Don't think we should allow Validators changing the shape at all.

Let's allow Contributors and Administrators to change the shape for Drafts, and just that for now. We can add more permissions in the future.

When the shape is changed after the initial submission, it should be logged in the activity log, but we don't have a clear way of logging shape diffs, so I'd rather not do it right now.

rajadain (changes requested):
This is generally looking good!

Let's take away the Validator's ability to change the shape.

Also, it would be nice to fix the map flashing. Would recommend that be given a shot in this PR, and only deferred if too complex.

src/django/api/views/boundary.py
mstone121: I was thinking reviews had their own shape, but after I checking the data model I realized they do not. [fea16bc](https://github.com/azavea/iow-boundary-tool/pull/141/commits/fea16bc559b68e9a83eee8a828fb08e61ff07e7c)

src/app/src/components/DrawTools/DrawTools.js
```diff
@@ -45,6 +76,18 @@ export default function DrawTools({ details }) {
     );
 }
 
+function getDrawMode({ status, userRole }) {
+    if (userRole === ROLES.VALIDATOR && status === BOUNDARY_STATUS.IN_REVIEW) {
```
mstone121, line 80: This used to be:
```js
        [BOUNDARY_STATUS.SUBMITTED, BOUNDARY_STATUS.IN_REVIEW].includes(
            details.status
        ) &&
        user.role === ROLES.VALIDATOR
```
I'm thinking that the distinction between `IN_REVIEW` and `SUBMITTED` is the presence of a row in the Reviews database table. If a Validator clicks "Review" on a boudary, an entry should be added to the Reviews table before taking them to the `/draw` page. 

Otherwise the Review creation logic would happen the first time a review is saved. This feels more complicated to me, however, it's something that could be handled entirely by the back-end.

I'm thinking that the distinction between `IN_REVIEW` and `SUBMITTED` is the presence of a row in the Reviews database table

src/app/src/pages/Draw.js
```diff
@@ -1,65 +1,18 @@
...
-    ) {
-        mode = DRAW_MODES.FULLY_EDITABLE;
-    }
-
     return (
         <Flex>
             <Sidebar />
             <Box flex={1} position='relative'>
                 <Map>
-                    <Layers mode={mode} />
```
mstone121, line 61: Is `Layers` going to need access to `mode`? If so, maybe boundary details should be loaded in a separate component above `DrawTools` and `Layers`, but below `Map`.

rajadain (changes requested):
Playing with this a bit, I think the reverting on fail to save is too harsh. For rural areas that may have flaky internet, a slight drop in connectivity can make the user lose their work:

https://user-images.githubusercontent.com/1430060/196787166-e2f56918-de21-4a0d-97d5-ed698376ac57.mp4

Failures here should be silent, and the shape on the map should remain unchanged.

When the contributor clicks "Save and back" or "Review and submit", we should also send the shape then. If there are internet connectivity issues then, we should let the user know, and stop them from navigating away from this page and losing their work. All this should be done as part of #132.

For this card, making the save failures silent in the background is enough.

Does that make sense?

src/app/src/components/DrawTools/DrawTools.js
rajadain: This understanding is correct.

DRAFT = Row in Submission table, with submitted_at = NULL
SUBMITTED = That row has a value for submitted_at
IN_REVIEW = Row in Review table, with reviewed_at = NULL
NEEDS REVISIONS = That row has a value for reviewed_at
APPROVED = Row in Approval table, with a value for approved_at

I'm not sure how much we'll end up using IN_REVIEW in practice.

src/app/src/pages/Draw.js
rajadain: I like that arrangement.

Alternatively, if `Layers` uses the same RTKQuery hook to get the boundary details, will it not get a cached value? i.e. it should be fetched only once for the different components that request it, right?

mstone121: > Alternatively, if Layers uses the same RTKQuery hook to get the boundary details, will it not get a cached value? i.e. it should be fetched only once for the different components that request it, right?

Yep, but then we would also need to handle the various states of the request (isFetching, error, success) in that component as well. I like to have one parent component handling all this and passing the non-null data to the children.

mstone121:
> Does that make sense?

Yep. See changes here: https://github.com/azavea/iow-boundary-tool/pull/141/commits/833ad59769812e9f4b1e3a16f4ae192c898005a3.

I'm not going to `fixup` this commit because it can serve as a refence for undoing manual cache updates.

src/django/api/views/boundary.py
```diff
@@ -63,3 +68,32 @@ def get(self, request, id, format=None):
...
+                ),
+            )
+
+        serializer = ShapeSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        boundary.latest_submission.shape = serializer.validated_data
+        boundary.latest_submission.save()
+
+        return Response()
```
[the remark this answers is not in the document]
mstone121: [c7f1035](https://github.com/azavea/iow-boundary-tool/pull/141/commits/c7f1035d7f2e6f3c5766122fe453989d6038b975)

src/app/src/hooks.js
```diff
@@ -110,3 +112,46 @@ export function useFilePicker(onChange) {
...
     return openFileDialog;
 }
+
+export function useBoundaryId() {
+    return useParams().boundaryId;
+}
+
+export function useTrailingDebounceCallback({
+    callback,
+    immediateCallback,
```
rajadain, line 122: Consider adding a comment here describing the expected difference between `callback` and `immediateCallback`

rajadain (approved):
+1 tested this is looking good! Nice work guiding this in through all the different reviews!

PR #142, opened 2022-10-19
Let users select utility at login

### LINKED ISSUE #124: Have Contributors Select Active Utility After Login
## Overview

After a user logs in, if they are a contributor, and they have access to more than one utility, they should see a page like this:

image

which should be populated with the utilities they have access to. After they select one, they should be taken to the Welcome page.

Their selected utility should be set as the currently active one.

If they have access to only one utility, this page should not be shown. Instead, that should be set as the active utility and they should go to the Welcome page directly.

If they have no utilities, they should see an error page that says that they need to be assigned a utility to proceed, with only the logout button as the only action they can perform.

If they have a draft with this selected utility, they should be taken to the Draw Page with that Draft loaded.

jacobtylerwalls:
## Overview

Allow contributors to select a utility when logging in (assuming there is more than one to choose from.)
Navigate directly to any drafts for that utility.

Closes #124 

### Demo

forthcoming

### Notes
- I didn't use RTK query to fetch the draft boundary in the "Proceed" button action. I tried, but was just burning time, so I quit. Feel free to suggest something more idiomatic. (We're not using RTK query for the rest of the login flow.)
- f9eab1aba8b05548b097f84c775e6ccc704998e6 adds a second test utility for the contributor user to make testing easier. I don't know if we want to merge it -- do we want staging to stay frictionless, i.e. don't have this form show?

## Testing Instructions

- scripts/resetdb
- Login as c1@azavea.com/password
  - [x] Ensure utility picker accepts your choice
  - [x] Navigates to the first draft
  - [ ] Or navigates to your location before login if there is no draft
- Repeat with a user with only one utility (just drop f9eab1aba8b05548b097f84c775e6ccc704998e6 and resetdb)
  - [x] ensure no selector
- Regression test utility selector in submissions page

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/pages/SelectUtility.js
+++ b/src/app/src/pages/SelectUtility.js
@@ -1,0 +1,63 @@
+import { Button, Text } from '@chakra-ui/react';
+import { useDispatch, useSelector } from 'react-redux';
+import { useNavigate } from 'react-router-dom';
+
+import apiClient from '../api/client';
+import LoginForm from '../components/LoginForm';
+import UtilityControl from '../components/UtilityControl';
+import { logout } from '../store/authSlice';
+
+import { API_URLS } from '../constants';
+
+export default function SelectUtility() {
+    const dispatch = useDispatch();
+    const navigate = useNavigate();
+    const locationBeforeAuth = useSelector(
+        state => state.auth.locationBeforeAuth
+    );
+    const destination = locationBeforeAuth || '/welcome';
+    const utilities = useSelector(state => state.auth.user?.utilities);
+    const width = '320px';
+
+    return utilities?.length ? (
+        <LoginForm>
+            <Text textStyle='loginHeader'>Boundary Sync</Text>
+            <Text textStyle='utilitySelectLabel' w={width} pt={2} pl={2}>
+                Select your utility
+            </Text>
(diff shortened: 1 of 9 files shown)


rajadain:
Taking a look now

src/app/src/pages/SelectUtility.js
```diff
@@ -0,0 +1,76 @@
...
+export default function SelectUtility() {
+    const dispatch = useDispatch();
+    const navigate = useNavigate();
+    const locationBeforeAuth = useSelector(
+        state => state.auth.locationBeforeAuth
+    );
+    const destination = locationBeforeAuth || '/welcome';
+    const utilities = useSelector(state => state.auth.user?.utilities);
+    const utility = useSelector(state => state.auth.utility);
+    const width = '320px';
```
rajadain, line 21: Consider defining this outside this component

```js
const UTILITY_CONTROL_WIDTH = '320px';
```

rajadain (approved):
This looks good. I'm fine with adding a second utility to c1. Alternatively, we could have a new contributor c2@azavea.com that has access to 2 utilities, and keep c1 with only one, so we could test both flows.

PR #144, opened 2022-10-20
Upload reference image metadata from draw tool

### LINKED ISSUE #126: Support Opacity for Reference Images
## Overview

At some point, we may allow the user to set an opacity value for reference images, beyond the one setting available in L.DistortableImage. In preparation, add an integer column to the ReferenceImage model to store the opacity value. It should be limited to values between 0 and 100. It should default to 100. The reference image update endpoint should allow setting this value, in addition to the distortion parameters.

### LINKED ISSUE #131: Draw Page: Add, Update Reference Images
## Overview

When a reference image is added in the Draw Page, it should be uploaded to the back-end.

Whenever a reference image is manipulated (distorted, made visible or not, or have its opacity adjusted), that should be updated in the back-end.

jacobtylerwalls:
## Overview
Upload and update reference image metadata from the draw page.

Closes #131
Closes #126

### Notes
- I took a second of learning time to see if I could reduce the size of the diff by providing [extraReducers](https://redux-toolkit.js.org/rtk-query/usage/examples#using-extrareducers) to `createSlice`, in the hopes of not having to manually update redux state, but IIUC the complexity tradeoff didn't look attractive because 1.) we have multiple existing actions (`toggleReferenceImageVisibility` and `updateReferenceImage`) and because the js and python parameters have to be mapped to each other anyway.
- Do we need to save `outlined` as part of this card?

## Testing Instructions

- scripts/migrate
- Use the draw tool to upload a reference image
  - [x] use a shell or the django admin to verify the file metadata was saved with the correct title
- Manipulate visibility, corners, transparency
  - [ ] use a shell or the django admin to verify the metadata was updated

## Checklist

- [ ] CI passes after rebase


jacobtylerwalls:
Ah hang on, a bit of error handling would be nice.

src/app/src/api/boundaries.js
```diff
@@ -48,6 +48,24 @@ const boundaryApi = api.injectEndpoints({
...
+        }),
+
+        updateReferenceImage: build.mutation({
+            query: updatedSubmission => ({
+                url: `/boundaries/${updatedSubmission.boundary}/reference-images/${updatedSubmission.id}/`,
+                method: 'PUT',
+                data: updatedSubmission,
+            }),
+            invalidatesTags: getUpdateItemTagInvalidator(TAGS.BOUNDARY),
+        }),
```
mstone121, line 68: A couple of things here:
- I don't think `newSubmission` makes as much sense here. Perhaps we can destructure some of these values.
- We shouldn't be using `BOUNDARY` tags for reference images unless we want them to be invalidated together. Since `query` has different arguments than boundaries, we also can't use the helper function for the update endpoint. Maybe at this point, we omit the tags and let the components re-fetch after they've been unmounted.
- There should be endpoints for querying the reference images. 
- I think there's enough going on here that reference images could have their own file in `/src/app/src/api`. 

```suggestion
        getReferenceImages: build.query({
            query: boundaryId => ({
                url: `/boundaries/${boundaryId}/reference-images/`,
                method: 'GET',
            }),
        }),

        uploadReferenceImage: build.mutation({
            query: ({ boundaryId, ...details }) => ({
                url: `/boundaries/${boundaryId}/reference-images/`,
                method: 'POST',
                data: details,
            }),
        }),

        updateReferenceImage: build.mutation({
            query: ({ boundaryId, referenceImageId, ...details }) => ({
                url: `/boundaries/${boundaryId}/reference-images/${referenceImageId}/`,
                method: 'PUT',
                data: details,
            }),
        }),
```

mstone121 (changes requested):
> I took a second of learning time to see if I could reduce the size of the diff by providing [extraReducers](https://redux-toolkit.js.org/rtk-query/usage/examples#using-extrareducers) to createSlice, in the hopes of not having to manually update redux state, but IIUC the complexity tradeoff didn't look attractive because 1.) we have multiple existing actions (toggleReferenceImageVisibility and updateReferenceImage) and because the js and python parameters have to be mapped to each other anyway.

I think what's adding to the complexity here is that there is not a single source of truth. The `mapSlice` reducer and the RTK Query API (and transitively the DB) are effectively storing much of the same information.

I think we should drop everything from the `mapSlice` reducer except for visibility, add a `query` endpoint for the reference image data and edit the reference image mutations to optimistically update the `query` endpoint.

We might want to debounce this endpoint too. https://github.com/azavea/iow-boundary-tool/pull/141 can serve as a reference for that. I'd be happy to pair tomorrow to go over it.

src/app/src/api/boundaries.js
jacobtylerwalls: Yeah, happy to have a look at this tomorrow. I was assuming we had to invalidate the boundary because we weren't going to have an endpoint for querying the images. So several things could change here.

jacobtylerwalls:
Thanks for the review, good ideas all around. So far I implemented:
- refactoring suggestions (destructuring, and new file in `src/api/...js`)
- removing `invalidatesTags`

I'm hoping for @rajadain's input on whether to:
- add a GET endpoint for reference-images (versus just loading them from the boundary details endpoint)
- debounce the reference image update endpoint for dragging corners (I'm concerned this is a nice to have that we might want to defer instead...)
- Whether saving the `outlined` prop to the backend is part of this card

rajadain (commented):
In response to the questions above:

1. What is the use case for a GETing a reference image? We should get their previous state when loading the boundary as we enter the `/draw/` page, and any further updates which are POSTed should return the next saved state.
2. We have an implementation of [`useTrailingDebounceCallback`](https://github.com/azavea/iow-boundary-tool/blob/fe5dc8ace993afb08e45b5bf2371e569ba0fa409/src/app/src/hooks.js#L120-L153) as of #141. Let's use that here.
3. It's likely that we're going to drop / hide the outline functionality in the near future, so let's ignore that for now.

src/app/src/components/Layers/ReferenceImageLayer.js
```diff
@@ -36,8 +52,26 @@ export default function ReferenceImageLayer() {
...
+    const updateReferenceImageDebounced = useTrailingDebounceCallback({
+        callback: (id, boundary, layer) => {
+            postReferenceImage(
+                convertImageMetadataFromStateFormat(id, boundary, layer)
+            );
+        },
+        immediateCallback: (id, boundary, layer) => {
+            dispatch(
+                api.util.updateQueryData(
+                    'updateReferenceImage',
```
mstone121, line 64: I think this should be `getBoundaryDetails` instead of `updateReferenceImage`. When the reference image is updated, we're changing what would be returned from `getBoundaryDetails`, so this is the cache that should be updated.

If we can update the `getBoundaryDetails` cache like this in "real time" (i.e. without waiting to query the django server), then the cache will operate effectively as the store for boundary details. The map slice then need only store ephemeral state related to user interface (e.g. visibility of reference layers, whether the UI is in add polygon mode). 

I realize I'm requesting a fairly significant change, but I think having a single source of truth will save us a lot of headache later on. Let me know if you want talk about it. Also looping in @rajadain for his input.

mstone121 (changes requested):
While we don't need endpoints for querying reference images, I still think we should use the RTK query cache of the `getBoundaryDetails` query as the single source of truth for reference image data. If we decide not to do this, I'll re-review as is.

See my comment on ReferenceImageLayer.js for a more detailed explanation.

rajadain (commented):
Since RTK Query is new to us, I think it's worth putting in the time to get it right. Let's have a huddle about this.

src/app/src/components/ModalSections/FileUpload.js
```diff
@@ -14,14 +14,14 @@ import { CloudUploadIcon } from '@heroicons/react/outline';
...
-import { useAddReferenceImage, useFilePicker } from '../../hooks';
+import { useFilePicker } from '../../hooks';
 import { useSelector } from 'react-redux';
 
 export default function FileUpload({ PreviousButton }) {
     const navigate = useNavigate();
-    const addReferenceImage = useAddReferenceImage();
 
-    const addFiles = newFiles => newFiles.forEach(addReferenceImage);
+    // TODO figure out reference image upload from this page
```
mstone121, line 23: This is addressed somewhat in #145.

src/app/src/components/Sidebar.js
```diff
@@ -52,14 +48,44 @@ export default function Sidebar() {
...
+        createReferenceImage({
+            boundaryId,
+            filename: file.name,
+            is_visible: true,
+            distortion: null,
+            opacity: 100,
+        });
+    };
+
+    // TODO support multiple files
```
mstone121, line 79: This might work as is, but I'm worried about manual cache updates and race conditions. I think we have two approaches here:
- Process uploads synchronously
- Update the api to support batch upload

src/app/src/components/Layers/ReferenceImageLayer.js
```diff
@@ -26,53 +19,29 @@ const convertCornerFromStateFormat = corner => ({
...
-        },
-        interval: 3000,
-    });
-
     const createLayer = useCallback(
-        ({ url, id, boundary, corners, mode }) => {
-            const layer = new L.distortableImageOverlay(url, {
+        ({ id, corners, mode }) => {
+            // TODO use reference image url here
+            const layer = new L.distortableImageOverlay('', {
```
mstone121, line 44: I wonder if we could have the django api return a url to the reference image in S3 and use that here.

mstone121:
@jacobtylerwalls @rajadain Ready for review. Note that my changes break the actual display of the reference images until #131 is implemented. To test, you can hard code a URL in ReferenceImageLayer.js.

jacobtylerwalls:
@mstone121 Did you mean #145 or #81? #131 is this card being closed here.

mstone121:
Oh, sorry: #81 is correct. It's the card where image file storage is added.

src/app/src/components/Layers/ReferenceImageLayer.js
rajadain: That is the plan. The API will return a signed URL that will allow temporary read access to the S3 resource, so the front-end can request the resource from there directly.

rajadain:
Taking a look now

src/app/src/api/referenceImages.js
```diff
@@ -0,0 +1,80 @@
...
+                    dispatch(
+                        api.util.updateQueryData(
+                            'getBoundaryDetails',
+                            boundaryId,
+                            draftDetails => {
+                                draftDetails.reference_images.push(newImage);
+                            }
+                        )
+                    );
+                } catch {}
```
rajadain, line 29: Should we be handling an error here?

src/app/src/api/referenceImages.js
```diff
@@ -0,0 +1,80 @@
...
+                            draftDetails.reference_images.map(image =>
+                                imageIsImageToBeUpdated(image)
+                                    ? { ...image, ...updatedImage }
+                                    : image
+                            );
+                    }
+                )
+            );
+        },
+        interval: 3000,
```
rajadain, line 71: Since we're using the same interval for the shape and the reference images, let's pull this into `constants.js`

src/app/src/api/referenceImages.js
```diff
@@ -0,0 +1,80 @@
...
+                api.util.updateQueryData(
+                    'getBoundaryDetails',
+                    boundaryId,
+                    draftDetails => {
+                        const imageIsImageToBeUpdated = image =>
+                            image.id === updatedImage.id;
+
+                        draftDetails.reference_images =
+                            draftDetails.reference_images.map(image =>
+                                imageIsImageToBeUpdated(image)
```
rajadain, line 63: Prefer a simpler name, like `shouldUpdate`

```suggestion
                                shouldUpdate(image)
```

rajadain (approved):
Great work, both of you! Left some minor comments.

This is definitely complex, and could be problematic if left unchecked. But it feels like an appropriate level of complexity for now, we'll keep an eye on it.

jacobtylerwalls:
Thanks both. Going to push some commits for review edits to keep this moving.

src/app/src/api/referenceImages.js
mstone121: It's not clear from [the documentation](https://redux-toolkit.js.org/rtk-query/api/createApi#onquerystarted) if this error will be the same error that is returned from the query function to the mutation hook, but in any case, that hook will still get the error message if there is one.

Here, we don't want to update the cache if the post request fails, so I think it's fine to not do anything.

jacobtylerwalls: I'll add a comment 👍🏻

mstone121 (approved):

PR #152, opened 2022-10-24
Add missing migration for User.full_name

jacobtylerwalls:
## Overview
Add a missing migration from #143.

## Testing Instructions

- scripts/manage resetdb
- `scripts/manage makemigrations` should not produce any further migrations

## Notes
Follow-up CI check in #153.

## Checklist

- [x] `fixup!` commits have been squashed
- [ ] `CHANGELOG.md` updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines
- [x] `README.md` updated if necessary to reflect the changes
- [x] CI passes after rebase

PR #153, opened 2022-10-24
Add CI check for missing migrations in `scripts/lint`

jacobtylerwalls:
## Overview
Add a step to `scripts/lint` that locally, produces missing migrations, and on CI, will exit with a failure code if migrations still need to be made.

### Demo
I expect this very run to fail until rebasing on #152 

### Notes
Omit `--dry-run`, which is often used to avoid actually creating the migrations. Here, like running `black`, just let them be made.

## Testing Instructions

- Make model changes without a migration
- scripts/lint

## Checklist

- [x] CI passes after rebase


mstone121 (approved):
Thanks for adding this. I always forget to run `makemigrations`.

PR #145, opened 2022-10-20
Wire up create boundary 

mstone121:
## Overview

This PR adds endpoints, serializers, and views for the File Upload/Create Boundary page.

Closes #129 

### Notes

Draft PR

## Checklist

- [ ] `fixup!` commits have been squashed
- [ ] `CHANGELOG.md` updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines
- [ ] `README.md` updated if necessary to reflect the changes
- [ ] CI passes after rebase

PR #154, opened 2022-10-25
Add draw page context

mstone121:
## Overview

This PR adds a react context for the draw page. This should make it easier to access boundary details from various components on the page.

## Checklist

- [ ] CI passes after rebase


src/app/src/components/DrawTools/EditToolbar.js
```diff
@@ -28,18 +28,22 @@ import {
...
-    const { polygon, editMode, addPolygonMode } = useSelector(
+    const boundary = useDrawBoundary();
+    const { editMode, addPolygonMode, polygonIsVisible } = useSelector(
         state => state.map
     );
 
     const confirmDeleteDialogController = useDialogController();
     const editDialogController = useDialogController();
 
+    const hasShape = Boolean(boundary.submission?.shape);
```
rajadain, line 45: Consider using something more idiomatic of JavaScript rather than a TypeScript convention.

```suggestion
    const hasShape = !!boundary.submission?.shape;
```

mstone121:
@rajadain The build is working locally for me after pushing updates, it introduced a bug with the polygon. I'll let you know when it's working again.

rajadain (commented):
I like this approach, am in favor. Will formally approve once the build passes. Will rebase my work for #96 on top of this once in.

mstone121:
@rajadain Ready for review. I had to also make some changes to reference layers after the rebase.

src/django/api/views/boundary.py
```diff
@@ -83,26 +83,46 @@ class BoundaryShapeView(APIView):
...
+        boundary.latest_submission.save()
+
+        return Response(status=HTTP_204_NO_CONTENT)
+
+    def delete(self, request, id, format=None):
+        boundary = BoundaryShapeView.get_boundary(request, id)
+        BoundaryShapeView.check_user_has_access_to_shape(request.user)
+        BoundaryShapeView.check_boundary_is_editable(boundary)
+
+        boundary.latest_submission.shape = None
```
rajadain, line 103: This fails because currently the shape is not nullable:

https://github.com/azavea/iow-boundary-tool/blob/38ca58c0685cb7c0988f68f7ec35258e369d0bb1/src/django/api/models/submission.py#L19

We can either make that field nullable, or we can make it so that deletion only happens in the front-end, and we write a new shape the next time the user creates / uploads one. I have a preference for the second approach, because ensuring that every submission has a shape simplifies things downstream.

rajadain (changes requested):
+1, nice work! I like how the context, the provider, and the usage hooks are grouped together. Also like seeing so much code be consolidated in one place, and this feels like a cleaner design.

Left a comment about better handling deletes, will take another look once that's been taken care of.

src/django/api/views/boundary.py
rajadain: Ah I see we're making that field nullable in #145. Once this is rebased on that it should start working correctly.

mstone121: I actually ran into this issue for different reasons in #145 and [made the field nullable there](https://github.com/azavea/iow-boundary-tool/pull/145/commits/9b089d8f8bf84bd3db29eac1af75106e9bb4b2c1#diff-c93adfd0016a9a82d8efb02214a793105a76ecc7c66f2f225ed8ddd73590f688). The migration is tied with other changes in a single commit. I suppose I could rebase and extract those changes into a separate commit if needed. 

I agree that ensuring every submission has a shape would be nice, but this could be a problem when creating a new boundary because we would need a default shape. And this shape would appear when a user first sees the draw page unless we add some logic to the draw page to check if this is a first-time-load, and if so, don't draw the polygon (or set it deleted in the front-end).

rajadain: You're right, I'd forgotten about creating a blank draft submission. Making it nullable is good 👍

rajadain (approved):
I like this approach. If #145 is taking longer to review and get in, consider pull in the bits that make Submission.shape nullable into this PR and merge this first.

PR #170, opened 2022-10-28
Authentication Fixes

### LINKED ISSUE #78: Use spinner to avoid flashing UI components before redirecting to login
## Overview
We could use a spinner (or a blank `<Loading>` component?) to avoid flashing the UI components such as the `/draw` page when logged out and the user is redirected to login.

### Additional Context
Cleanup identified in https://github.com/azavea/iow-boundary-tool/pull/77#discussion_r979419979.

### LINKED ISSUE #149: API requests happen before auth check is complete.
## Overview

API requests are being sent before authentication checks complete.

This is important for two reasons:
- We don't want to show data to users who shouldn't see it.
- We want the correct data to load for each user. The auth check returns info about the user which determines what shows up on the page.

This is probably because the RTK query hooks used in components to make API requests are fired as soon as they are mounted. Rather than trying to tell each component to wait until auth, we should probably make it so these components don't mount until the auth check completes. Implementing this could also resolve #78.

### Expected Behavior

API requests should be sent after auth checks. 

### Actual Behavior

API requests happen are being sent before auth checks.

### Steps to Reproduce

Note: These steps assume #148 has been merged.

- Login as a c1@example.com
- Go to the submissions page
- Open the network tab
- Refresh the page from the browser
- Notice the request to `/api/boundaries` with no query parameters
- Select a different utility
- Notice the request to `/api/boundaries/?utilities=<selected utility id>`
- Select the original utility
- Notice the request to `api/boundaries/?utilities=<original utility id>`

### Demo

https://user-images.githubusercontent.com/8356789/197547066-f5788416-db38-4817-89cd-2fa31c12ffb9.mov

### LINKED ISSUE #164: QA the utility selector for validators and administrator users
## Overview

Demos showed that the app might not handle gracefully validators and administrators having no selected utility when navigating among activities, such as to and from the welcome tool. This indicates a need to QA this tool more.

### Steps to Reproduce
TBD

### LINKED ISSUE #168: Validators should not be able to go to Welcome Page
## Overview

They should be redirected to the Submissions List.

mstone121:
## Overview

This PR does mainly three things:
- Prevents the app from rendering without an authenticated user
- Prevents non-contributors from getting to the welcome page
- Prevents contributors from accessing the app without selecting a utility

Closes #168 
Closes #78 
Closes #149 
Closes #164 

### Demo

https://user-images.githubusercontent.com/8356789/198731408-957e1812-71de-4d34-b3ba-63c45d74085b.mov

### Notes

The private routes are moved into an auth guard which prevents render of children until the user is authenticated.

The utility guard might need to moved into the private routes element and surround only certain elements. For example, if a user visits a submission details page, they'll still have to select a utility. 

## Testing Instructions

- Go to http://localhost:4545
- Log in
- Navigate around
- Refresh a page, a center spinner should be visible breifly before the app loads
- You should not have had to log in again.
- Log out
- Go to a private page
- You should be redirected to the login page

## Checklist

- [x] CI passes after rebase


rajadain:
Taking a look now

rajadain (changes requested):
This seems to break the "redirect to originally requested private route after login" functionality, as demoed here:

https://user-images.githubusercontent.com/1430060/198903426-c2d25afc-01c6-4ac5-aa7c-fb2745f1bd7e.mp4

I should be redirected to `/submissions/4`, but instead I go to `/welcome` and then `/draw/1`.

mstone121:
> This seems to break the "redirect to originally requested private route after login" functionality, as demoed here:

The code to fix this ended up being mixed with the utility selector so I assigned myself that and included the fixes here. I added a utility guard much like the auth guard (but that does not redirect).

We might need to think about moving this further into the route tree. If a user visits a submission details page, they'll still have to select a utility. Perhaps if they visit a submission details page, we should auto select the utility based on the submission.

rajadain:
Taking another look. I made #172 as a follow-up which may need the `NavBar` to be moved to be within the BoundaryProvider.

rajadain (approved):
+1 tested, nice work on this!

mstone121:
> Taking another look. I made #172 as a follow-up which may need the `NavBar` to be moved to be within the BoundaryProvider.

Another possibility to move the utility guard into [PrivateRoutes](https://github.com/azavea/iow-boundary-tool/blob/8783b0321ea3452cde5a4e049af56635bd6f807c/src/app/src/App.js#L52) so that it encapsulates only the routes that have multiple different boundary possibilities. The others will set the utility themselves.

PR #156, opened 2022-10-26
Add User Permissions per Boundary, Guard Draw Page Actions

rajadain:
## Overview

Adds a `useDrawPermissions` hook which returns if the current user `canWrite`, `canReview`, or `canApprove` the current boundary. Guards actions for uploading reference images, editing the shape, to only when the user `canWrite`. Shows appropriate navigation button in the Boundary Detail page based on permissions.

Also fixes an issue where hitting back from the Draw page after getting there from the Boundary Detail would take the user to the Boundary List instead of the Boundary Detail.

Also removes the extraneous Save and Back button on the Draw/Map, since now we have one in the NavBar.

Also centers the boundary on Draw Page load.

Builds atop #154
Closes #96 
Closes #166

### Demo

https://user-images.githubusercontent.com/1430060/198140019-bdeb5806-3756-4190-b854-62d94ea7b10b.mp4

### Notes

Should we make the `DrawContext` a `BoundaryContext` and have it be available to Boundary Detail Page as well?

## Testing Instructions

- Check out this branch and `update` and `resetdb` and `server`
- Go to http://localhost:4545/ and login as c1@azavea.com
  - [x] Ensure you can edit the Draft
  - [x] Ensure you can view, but not edit the other submissions
- Log out, log back in as v1@azavea.com
  - [x] Ensure you can Start a Review for the submitted boundary
  - [x] Ensure you can Continue a Review for the in review boundary
  - [x] Ensure you see an Approve button for both boundaries
  - [x] Ensure you can view, but not edit any submission

## Checklist

- [x] `fixup!` commits have been squashed
- [x] `CHANGELOG.md` updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines
- [x] `README.md` updated if necessary to reflect the changes
- [x] CI passes after rebase

PR #173, opened 2022-10-31
Handle boundaries with no shape

### LINKED ISSUE #163: Handle Boundaries with No Shapes
## Overview

When a new boundary is created, and a draft submission for it, it initially has no shape. A lot of the Draw Page relies on that shape being there.

We should discourage users from navigating away from the Draw Page when there is no shape:

- [ ] Disable the "Review and submit", "Save and back" buttons while there is no shape

But this will not be enough, because the users can still close the tab, leaving no shape attached to the boundary. To handle this case, we have to update the Draw Page (and Boundary Detail Map) to work without a shape:

- [ ] Allow for a boundary not having a shape when loading the Draw Page for an existing boundary
- [ ] Allow for a boundary not having a shape when loading the Boundary Detail page

mstone121:
## Overview

A boundary initially has no shape. This PR updates the UI to handle this situation.
- Review and Submit button is disable with no boundary
- On the submission details page, if there is no shape for the submission, the map will be empty and there will be no download button.

This PR also fixes an issue where an added and unmodified boundary makes the details page unloadable.

Closes #163 

### Notes

#163 requests that the "Save and back" button be disabled. This was renamed to just "Back" #156 so the NavBar wouldn't need to know about the current boundary. Disabling it here would require the same functionality so this change is not included. 

## Testing Instructions

- http://localhost:4545
- Login as c1@azavea.com
- Select "Other utility" as utility
- Don't draw a shape
- The "Review and Submit" button should be disabled.
- Refreshing the page should return the draw page to the same state
- Clicking back takes you to the details page without error
- There should be no download button
- Click edit boundary
- Draw a boundary
- Refresh page and check that everything loads correctly

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Wire up create boundary [#145](https://github.com/azavea/iow-boundary-tool/pull/145)
 - Add Draw page React context [#154](https://github.com/azavea/iow-boundary-tool/pull/154)
 - Add User Permissions per Boundary [#156](https://github.com/azavea/iow-boundary-tool/pull/156)
+- Handle boundaries with no shape [#173](https://github.com/azavea/iow-boundary-tool/pull/173)
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


rajadain:
Taking a look

rajadain (approved):
+1 tested, nice work!

PR #175, opened 2022-11-02
Add Support for Uploading Files

### LINKED ISSUE #81: Add file fields to Reference_Image and Submission

### LINKED ISSUE #158: Wire up front-end to upload files
## Overview

Once the file upload APIs are available in #81, wire up the front-end to upload files to the back-end.

Files should be uploadable from the Welcome page and the Draw page.

In the Welcome page, users should be able to select multiple kinds of files (images for reference, shapefiles for the boundary) from one file selection:

image

If they select more than one shapefile, we should prompt them that they they can have only one, and show a popup that lets them select one from the list of shapefiles they picked, or to pick anew.

In the Draw page, the reference image file picker should only allow selecting images. The boundary upload file picker should only allow selecting shapefiles, and only one at a time.

Uploaded shapefiles should be added to the map and saved to the database.

rajadain:
## Overview

Adds back-end support for uploading and saving files.

~~All uploaded files are saved to S3. It was easier to do this than to have development save locally. The development bucket will be shared between all developers and test environments.~~

Uploaded files are saved to local disk in development, and in S3 buckets in staging and production.

Supports multiple reference image uploads, and a single shape file upload. Supports uploading a zipped shapefile, as well as a geojson file.

Reference images are reported as signed S3 URLs, so the front-end can access them despite the buckets themselves being private. These URLs are valid for about an hour.

The front-end is updated to upload the files correctly, and also to load the shapes and reference images from the back-end. An edge case with L.DistortableImage was handled, without which navigating away from the map was not possible.

Also refactors the tests, updates the reference image tests to upload files, and adds tests for shapefile and geojson uploads.

~~More UI polish will come with #158.~~

Closes #81 
Closes #158 

### Demo

https://user-images.githubusercontent.com/1430060/199588222-5961e7e9-6963-4288-9b6e-26c1c1f88d4b.mp4

## Testing Instructions

- Check out this PR, run `update` and `resetdb` and `server`
- Log in as a contributor, go through the Welcome modals
- Select a number of files for uploading
  - ℹ️ Use the shapefile or geojson added for testing if you don't have any locally
  - ℹ️ Use any images available for testing reference images
  - [x] Ensure that the uploads are handled correctly
  - [x] Ensure that the reference images are rendered correctly
  - [ ] Ensure that the shapefile / geojson is ingested correctly
- Navigate around the app
  - [ ] Make sure there's no other crashes

## Checklist

- [ ] CI passes after rebase


mstone121 (changes requested):
I'm seeing this error when I try to upload:
Screen Shot 2022-11-03 at 9 35 02 AM
Do I need to do some AWS config? If so, maybe we need to add some readme instructions.

Also, the app appears to be crashing after `./scripts/resetdb` and visiting `/draw/1` with 'Other Utility'.

rajadain:
Removed S3 requirement for development in [12d6847](https://github.com/azavea/iow-boundary-tool/pull/175/commits/12d6847232b6c1542bd346cc79829916ee3ad9f1)

Added a few commits [d7a8e29](https://github.com/azavea/iow-boundary-tool/pull/175/commits/d7a8e29672b98ec594d5cdf545a96adf9dc07d4f), [23c8599](https://github.com/azavea/iow-boundary-tool/pull/175/commits/23c8599345f8f392bd004d7f304e7d7a5f9add7b), [63da524](https://github.com/azavea/iow-boundary-tool/pull/175/commits/63da52410c09d9192b10db21355a75294439f617), [5af0f1c](https://github.com/azavea/iow-boundary-tool/pull/175/commits/5af0f1cb28f9dff72c7b8973ee747268f5076996), [a6753dc](https://github.com/azavea/iow-boundary-tool/pull/175/commits/a6753dc2241944dc48d5d790f98155f4d162ff9d) to complete the file upload experience in the front-end. This now also closes #158.

> Also, the app appears to be crashing after `./scripts/resetdb` and visiting `/draw/1` with 'Other Utility'.

I'm not able to reproduce this, could have been related to the AWS_PROFILE issue, which should no longer be the case. Let me know if this is still happening.

This is ready for another look.

rajadain:
Rebased on latest `develop` to resolve merge conflicts

rajadain:
Ah I'm seeing the `draw/1` bug now, will investigate.

src/app/src/api/boundaries.js
```diff
@@ -75,10 +81,55 @@ const boundaryApi = api.injectEndpoints({
...
+                    data,
+                };
+            },
+            onQueryStarted: async ({ id }, { dispatch, queryFulfilled }) => {
+                const patchResult = dispatch(
+                    api.util.updateQueryData(
+                        'getBoundaryDetails',
+                        id,
+                        draftDetails => {
+                            draftDetails.submission.shape = null;
```
mstone121, line 108: Why delete the shape before we know the patch succeeds?

src/app/src/components/DrawTools/EditPolygonModal.js
```diff
@@ -14,15 +14,30 @@ import {
...
-export default function EditPolygonModal({ isOpen, defaultLabel, onClose }) {
-    const [label, setLabel] = useState(defaultLabel);
+import { ACCEPT_SHAPES } from '../../constants';
+import { useBoundaryId, useFilePicker } from '../../hooks';
+import { useReplaceBoundaryShapeMutation } from '../../api/boundaries';
 
-    useEffect(() => setLabel(defaultLabel), [defaultLabel]);
+export default function EditPolygonModal({ isOpen, label, onClose }) {
+    const [isLoading, setIsLoading] = useState(false);
+    const [replaceShape] = useReplaceBoundaryShapeMutation();
```
mstone121, line 23: Mutations return meta data about the request as the second object in the returned tuple. Among these data is `isLoading`:
```js
 const [replaceShape, {isLoading}] = useReplaceBoundaryShapeMutation();
```

src/app/src/components/ModalSections/FileUpload.js
```diff
@@ -22,11 +23,13 @@ import {
 } from '../../utils';
 import { useEndpointToastError, useFilePicker } from '../../hooks';
 import { useStartNewBoundaryMutation } from '../../api/boundaries';
+import { ACCEPT_BOTH } from '../../constants';
 
 export default function FileUpload({ PreviousButton }) {
     const toast = useToast();
     const navigate = useNavigate();
     const utilityId = useSelector(state => state.auth.utility.id);
+    const [isLoading, setIsLoading] = useState(false);
```
mstone121, line 32: I recommend using the mutation's `isLoading` value as described in a previous comment.

src/django/api/tests/reference_image.py
```diff
@@ -0,0 +1,88 @@
...
+        updated_image_illegitimate = {
+            **self.get_update_payload(),
+            "uploaded_by": self.administrator.pk,
+        }
+
+        response = self.client.put(
+            update_url, updated_image_illegitimate, format="json"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
mstone121, line 63: Double checking we don't want a `400 Bad Request` here

src/django/iow/settings.py
```diff
@@ -14,6 +14,7 @@
 import requests
 
 from pathlib import Path
+from django.core.exceptions import ImproperlyConfigured
```
mstone121, line 17: Is this import used?

mstone121 (changes requested):
Requesting a few changes. In addition to the code comments:
- I'm finding that uploading a shape from the welcome screen won't override an existing shape. 
- Would you mind updating the PR text to reflect that local file storage is being used in development?

src/django/api/tests/reference_image.py
rajadain: Read Only fields are ignored, as stated in the documentation: https://www.django-rest-framework.org/api-guide/fields/#read_only, so the presence of the read only key does not result in a 400. This design decision is debatable, but in the interest of efficiency I'm going to leave it as is.

src/django/iow/settings.py
rajadain: Yes, it was used without importing it:

https://github.com/azavea/iow-boundary-tool/blob/5e23e9713eaa18953b5f99a318b832b2d5684e1b/src/django/iow/settings.py#L30-L32

https://github.com/azavea/iow-boundary-tool/blob/5e23e9713eaa18953b5f99a318b832b2d5684e1b/src/django/iow/settings.py#L64

https://github.com/azavea/iow-boundary-tool/blob/5e23e9713eaa18953b5f99a318b832b2d5684e1b/src/django/iow/settings.py#L69-L70

This adds the missing import.

src/app/src/components/DrawTools/EditPolygonModal.js
rajadain: [0febfad](https://github.com/azavea/iow-boundary-tool/pull/175/commits/0febfada8f8e04cd2dabbf7902bf9c6ba8eb9943)

src/app/src/components/ModalSections/FileUpload.js
rajadain: [b2bc044](https://github.com/azavea/iow-boundary-tool/pull/175/commits/b2bc04431e7916c7aec85b852c73150a169fbfe8)

src/app/src/api/boundaries.js
rajadain: This is done mainly for this:

https://github.com/azavea/iow-boundary-tool/blob/5e23e9713eaa18953b5f99a318b832b2d5684e1b/src/app/src/components/DrawTools/useEditingPolygon.js#L131-L137

We want the map to zoom to the new shape once it is loaded. If we set `hasZoomedToShape` to `false`, that `useEffect` gets called immediately and zooms to the old shape. In order for it to not fire until the new shape is loaded, we first remove the old shape.

In case the patch does not succeed, we restore the old shape in the `catch` block:

https://github.com/azavea/iow-boundary-tool/blob/b2bc04431e7916c7aec85b852c73150a169fbfe8/src/app/src/api/boundaries.js#L126-L129

rajadain:
> I'm finding that uploading a shape from the welcome screen won't override an existing shape.

This is a good catch, although the solution may be almost out of scope and may require another card to solve properly. In the interest of budget I've included two commits: [a37fb01](https://github.com/azavea/iow-boundary-tool/pull/175/commits/a37fb019456bab1da427d108a61d11581910e2f5), [cc79043](https://github.com/azavea/iow-boundary-tool/pull/175/commits/cc790438c0ac2b8f1c2fff0f05243f4d5f913e86), and [99493d4](https://github.com/azavea/iow-boundary-tool/pull/175/commits/99493d4cd01b394a2d4d23a9e826e0f8cee0b7fe) which make it so that the user is not taken to the `/welcome` page unless there are no existing boundaries for the selected utility.

It doesn't make sense for a contributor to go through the `/welcome` page if they already have a boundary in the system.

I started writing a card for this, but realized the code change would not be that complex, so decided to include them here. I can also pull those into a separate PR if they feel too unrelated.

rajadain:
This is ready for another look.

src/app/src/api/boundaries.js
mstone121: What if we moved the `dispatch(setHasZoomeToShape(false))` call to after the `queryFulfilled` promise has been resolved and the shape has been updated?

mstone121 (approved):
Looks good!

src/app/src/api/boundaries.js
rajadain: That is what I had done originally, but it fires too quick, and still zooms to the old shape rather than the new one. Even when it is after the `dispatch(api.util.updateQueryData(...))` call.

mstone121: That makes sense. There might be a way to figure this out with async actions/extra reducers, but probably not worth getting into considering it works as is.

mstone121:
> This is a good catch, although the solution may be almost out of scope and may require another card to solve properly.

Just to clarify, is the solution you're referring to not what you implemented in the last three commits?

rajadain:
> Just to clarify, is the solution you're referring to not what you implemented in the last three commits?

I think these three commits are good for now, but there may be aspects of this workflow that need further polish which we may discover once these are in. Will make a follow up card when those issues are discovered.

rajadain:
Thanks for the thorough review! Will merge once green.

PR #179, opened 2022-11-04
Provide pre-commit hook to run format script

jacobtylerwalls:
## Overview
- Create a format script that has black and isort auto format python files
- Provide a pre-commit hook to run the format script

### Demo
Updated demo:

https://user-images.githubusercontent.com/38668450/200062598-d9021a81-c6fc-4445-b3a6-880d80139680.mp4

### Notes
- ~`pre-commit` could be specified in a `requirements-dev.txt`, and the README edited to just suggest installing _that_, but I figured with just one package, to suggest installing it directly.~ Got this in `scripts/bootstrap` now.

## Testing Instructions

- `scripts/bootstrap`
- Make changes to js and python that will cause lint failures, or edit a Django model to necessitate a migration
- Attempt to commit
  - [ ] verify you are prevented from committing

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,9 @@ Municipal data can be downloaded with a script:
 ./scripts/fetch-data
 ```
 
+A pre-commit hook that runs the format script before any commit will be
+installed but can always be ignored with `git commit --no-verify`.
+
 #### Development
 
 Rebuild Docker images and run application.
--- a/scripts/format
+++ b/scripts/format
@@ -20,6 +20,11 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     if [ "${1:-}" = "--help" ]; then
         usage
     else
+        # Format JavaScript app
+        docker-compose \
+            run --rm --no-TTY --entrypoint yarn app \
+            prettier --loglevel warn --write src/
+
         # Format Django app
 
         # Get a list of changed ".py" files and trim "src/django"
--- a/scripts/format
+++ b/scripts/format
@@ -30,26 +30,9 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         # Get a list of changed ".py" files and trim "src/django"
         changed_files=$(git status --porcelain | grep ".py" | awk -F'src/django/' '{print $2}' | xargs)
         if [[ -n "${changed_files}" ]]; then
-            # Run isort --check-only before fixing to ensure a failure code is emitted
-            if ! docker-compose \
-                run --rm --no-TTY --no-deps --entrypoint isort django ${changed_files} --check-only; then
-
-                docker-compose \
-                    run --rm --no-TTY --no-deps --entrypoint isort django ${changed_files}
-            fi
-            # Run black with --check before fixing to ensure a failure code is emitted
-            if ! docker-compose \
-                run --rm --no-TTY --no-deps --entrypoint black django \
-                --skip-string-normalization \
-                --extend-exclude migrations ${changed_files} \
-                --quiet --check; then
-
-                docker-compose \
-                    run --rm --no-TTY --no-deps --entrypoint black django \
-                    --skip-string-normalization \
-                    --extend-exclude migrations \
-                    --quiet ${changed_files}
-            fi
+            docker-compose \
+                run --rm --no-TTY --no-deps --entrypoint ./format.sh django \
+                ${changed_files}
         fi
     fi
 fi
--- a/scripts/format
+++ b/scripts/format
@@ -1,0 +1,50 @@
+#!/bin/bash
+# Disable "double quote to prevent globbing and word splitting"
+# ${changed_files} is a list that needs to be split on words 
+# shellcheck disable=SC2086
+
+set -e
+
+if [[ -n "${IOW_BOUNDARY_TOOL_DEBUG}" ]]; then
+    set -x
+fi
+
+function usage() {
+    echo -n "Usage: $(basename "$0")
+
+Run formatters on the project's code
+"
+}
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
+    if [ "${1:-}" = "--help" ]; then
+        usage
+    else
+        # Format Django app
+
+        # Get a list of changed ".py" files and trim "src/django"
+        changed_files=$(git status --porcelain | grep ".py" | awk -F'src/django/' '{print $2}' | xargs)
+        if [[ -n "${changed_files}" ]]; then
+            # Run isort --check-only before fixing to ensure a failure code is emitted
+            if ! docker-compose \
+                run --rm --no-TTY --no-deps --entrypoint isort django ${changed_files} --check-only; then
+
+                docker-compose \
+                    run --rm --no-TTY --no-deps --entrypoint isort django ${changed_files}
+            fi
+            # Run black with --check before fixing to ensure a failure code is emitted
+            if ! docker-compose \
+                run --rm --no-TTY --no-deps --entrypoint black django \
+                --skip-string-normalization \
+                --extend-exclude migrations ${changed_files} \
+                --quiet --check; then
+
+                docker-compose \
+                    run --rm --no-TTY --no-deps --entrypoint black django \
+                    --skip-string-normalization \
+                    --extend-exclude migrations \
+                    --quiet ${changed_files}
+            fi
+        fi
+    fi
+fi
--- a/scripts/lint
+++ b/scripts/lint
@@ -24,6 +24,9 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
 
         # Lint JavaScript
         docker-compose \
+            run --rm --entrypoint yarn app prettier --check src/
+
+        docker-compose \
             run --rm --entrypoint ./node_modules/.bin/eslint \
             app src/ --ext .js --ext .jsx
 
--- a/scripts/lint
+++ b/scripts/lint
@@ -54,13 +54,16 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
                 --exclude settings.py,manage.py,*.pyc,api/migrations/*
 
             docker-compose \
+                run --rm --no-deps --entrypoint isort django . --check
+
+            docker-compose \
                 run --rm --no-deps --entrypoint black django \
                 --skip-string-normalization --check --diff \
                 --extend-exclude migrations .
 
             docker-compose \
                 run --rm --entrypoint python django \
-                manage.py makemigrations
+                manage.py makemigrations --check --dry-run
         fi
     fi
 fi
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,7 @@
+repos:
+- repo: local
+  hooks:
+    - id: format
+      name: format
+      entry: scripts/format
+      language: script
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for uploading files [#175](https://github.com/azavea/iow-boundary-tool/pull/175)
 - Add unapproving to data model [#178](https://github.com/azavea/iow-boundary-tool/pull/178)
 - Add ability to submit reviews [#181](https://github.com/azavea/iow-boundary-tool/pull/181)
+- Add format script and pre-commit hook [#179](https://github.com/azavea/iow-boundary-tool/pull/179)
 - Add ability to resubmit boundaries [#185](https://github.com/azavea/iow-boundary-tool/pull/185)
 - Add support for approving/unapproving boundaries [#186](https://github.com/azavea/iow-boundary-tool/pull/186)
 
--- a/scripts/bootstrap
+++ b/scripts/bootstrap
@@ -11,7 +11,7 @@ DIR="$(dirname "${0}")/../"
 function usage() {
     echo -n \
         "Usage: $(basename "$0")
-Update environment variables file.
+Update environment variables file and install pre-commit hooks.
 "
 }
 
@@ -27,5 +27,10 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         fi
 
         popd
+
+        if [[ -z "${CI}" ]]; then
+            python3 -m pip install pre-commit
+            pre-commit install
+        fi
     fi
 fi
--- a/src/app/.prettierrc
+++ b/src/app/.prettierrc
@@ -1,10 +1,18 @@
 {
     "arrowParens": "avoid",
     "bracketSpacing": true,
+    "bracketSameLine": false,
     "htmlWhitespaceSensitivity": "css",
     "insertPragma": false,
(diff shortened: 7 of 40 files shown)


README.md
```diff
@@ -35,6 +35,18 @@ Municipal data can be downloaded with a script:
 ./scripts/fetch-data
 ```
 
+Install the pre-commit hook that runs the lint script before any commit.
+(`pre-commit` is a python package that facilitates installing pre-commit
+hooks; thus it runs on the host machine, not Docker, and is not included
+in `requirements.txt`.)
+
+```console
+pip install pre-commit
```
rajadain, line 44: Consider adding the pre-commit installation to `scripts/bootstrap`, to streamline setup.

We should still talk about it in the README though, and mention the `git commit --no-verify` escape.

jacobtylerwalls: [0a3b98a](https://github.com/azavea/iow-boundary-tool/pull/179/commits/0a3b98a99dc8837d5eca719e7b1e23fe0aff1f92)

jacobtylerwalls:
`isort` and `black` are formatters, but we were already running `black` in `lint`. I didn't see harm in continuing this usage, even though it fuzzes the boundary between linting and formatting.

jacobtylerwalls:
Running faster now that I've separated the lint and formatting concerns. This does mean that eslint and shellcheck won't run on pre-commit, which might be alright, as we're trying to keep the time down. What do you think?

rajadain:
I think that's alright.

eslint should be running continuously via `server`, so developers get feedback there. It is also a part of CI, so will ensure nothing breaking gets in.

shellcheck doesn't need to run as often because the shell files are rarely changed. It is also run as part of CI, so will ensure nothing breaking gets in.

Will take another look.

rajadain:
Heads up that this needs to be rebased as of #175

jacobtylerwalls:
I'll need to fix the way black is launched on changed files. Black doesn't take multiple files specified by name.

scripts/lint
```diff
@@ -53,14 +53,17 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
                 run --rm --no-deps --entrypoint flake8 django \
                 --exclude settings.py,manage.py,*.pyc,api/migrations/*
 
+            docker-compose \
+                run --rm --no-deps --entrypoint isort django . --check
+
```
rajadain, line 58: Consider also adding a lint check for prettier:

```diff
diff --git a/scripts/lint b/scripts/lint
index b1e23fc..b130574 100755
--- a/scripts/lint
+++ b/scripts/lint
@@ -23,6 +23,9 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
             scripts/*
 
         # Lint JavaScript
+        docker-compose \
+            run --rm --entrypoint yarn app prettier --check src/
+
         docker-compose \
             run --rm --entrypoint ./node_modules/.bin/eslint \
             app src/ --ext .js --ext .jsx
```

scripts/format
```diff
@@ -0,0 +1,34 @@
...
+
+Run formatters on the project's code
+"
+}
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
+    if [ "${1:-}" = "--help" ]; then
+        usage
+    else
+        # Format Django app
```
rajadain, line 20: Consider also adding prettier formatting for the JavaScript side

```diff
diff --git a/scripts/format b/scripts/format
index 411bb4e..0a44129 100755
--- a/scripts/format
+++ b/scripts/format
@@ -17,6 +17,10 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     if [ "${1:-}" = "--help" ]; then
         usage
     else
+        # Format JavaScript app
+        docker-compose \
+            run --rm --entrypoint yarn app prettier --write src/
+
         # Format Django app
         docker-compose \
             run --rm -T --no-deps --entrypoint isort django .
```

rajadain (commented):
Also discovered that to run prettier from the command line correctly, needed to change the config a little:

```diff
diff --git a/src/app/.prettierrc b/src/app/.prettierrc
index e519770..950c7bf 100644
--- a/src/app/.prettierrc
+++ b/src/app/.prettierrc
@@ -3,8 +3,15 @@
     "bracketSpacing": true,
     "htmlWhitespaceSensitivity": "css",
     "insertPragma": false,
-    "bracketSameLine": false,
     "jsxSingleQuote": true,
+    "overrides": [
+        {
+            "files": "*.css",
+            "options": {
+                "parser": "css"
+            }
+        }
+    ],
     "parser": "babel",
     "printWidth": 80,
     "proseWrap": "always",
```

The removal of `jsxBracketSameLine` is because that has been deprecated https://prettier.io/blog/2021/09/09/2.4.0.html.

The addition of `parser: css` is because by default it's using some other parser (unclear what) for CSS files https://stackoverflow.com/a/55614622/6995854.

jacobtylerwalls:
I think I found a solution that doesn't dramatically impact performance that allows us to check individual python files, format them automatically, and prevent the commit from going through. See [af06856](https://github.com/azavea/iow-boundary-tool/pull/179/commits/af06856da7204903f376598249def68fc279ad09) in particular.

scripts/format
```diff
@@ -0,0 +1,52 @@
...
+"
+}
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
+    if [ "${1:-}" = "--help" ]; then
+        usage
+    else
+        # Format JavaScript app
+        docker-compose \
+            run --rm --no-TTY --entrypoint yarn app prettier --write src/
```
rajadain, line 22:
```suggestion
            run --rm --no-TTY --entrypoint yarn app prettier --loglevel warn --write src/
```

rajadain (changes requested):
If I have multiple issues in multiple Python files, like this:

```diff
diff --git a/src/django/api/models/boundary.py b/src/django/api/models/boundary.py
index e7ba76d..0e70d7c 100644
--- a/src/django/api/models/boundary.py
+++ b/src/django/api/models/boundary.py
@@ -34,8 +34,7 @@ class Boundary(models.Model):
             return self.latest_submission.updated_at
 
         if self.status == BOUNDARY_STATUS.SUBMITTED:
-            if (
-                self.latest_submission.latest_approval is not None
+            if (self.latest_submission.latest_approval is not None
                 and self.latest_submission.latest_approval.revoked
             ):
                 return self.latest_submission.latest_approval.unapproved_at
diff --git a/src/django/api/models/reference_image.py b/src/django/api/models/reference_image.py
index 40c2a39..5436dba 100644
--- a/src/django/api/models/reference_image.py
+++ b/src/django/api/models/reference_image.py
@@ -13,8 +13,7 @@ def reference_image_upload_path(instance, filename):
 
 class ReferenceImage(models.Model):
     filename = models.CharField(max_length=255)
-    boundary = models.ForeignKey(
-        Boundary, on_delete=models.PROTECT, related_name='reference_images'
+    boundary = models.ForeignKey(Boundary, on_delete=models.PROTECT, related_name='reference_images'
     )
     uploaded_at = models.DateTimeField(auto_now_add=True)
     uploaded_by = models.ForeignKey(User, on_delete=models.PROTECT)
diff --git a/src/django/api/models/state.py b/src/django/api/models/state.py
index 1af1ad3..1d88d2c 100644
--- a/src/django/api/models/state.py
+++ b/src/django/api/models/state.py
@@ -9,8 +9,7 @@ class State(models.Model):
     id = CICharField(max_length=2, primary_key=True, unique=True)
     name = models.CharField(max_length=127)
     shape = gis_models.MultiPolygonField(geography=True)
-    options = models.JSONField(
-        blank=True,
+    options = models.JSONField(blank=True,
         null=True,
         help_text='This JSON dictionary contains state-specific configuration.',
     )
diff --git a/src/django/api/models/submission.py b/src/django/api/models/submission.py
index 7a2a8bf..e025230 100644
--- a/src/django/api/models/submission.py
+++ b/src/django/api/models/submission.py
@@ -16,8 +16,7 @@ def limit_by_validator_or_admin():
 
 
 def shape_upload_path(instance, filename):
-    return (
-        f"boundaries/{instance.boundary.id}/submissions/{instance.id}/shape/{filename}"
+    return (f"boundaries/{instance.boundary.id}/submissions/{instance.id}/shape/{filename}"
     )
 
 
diff --git a/src/django/api/models/user.py b/src/django/api/models/user.py
index cfd5d22..cb85fe3 100644
--- a/src/django/api/models/user.py
+++ b/src/django/api/models/user.py
@@ -77,8 +77,7 @@ class User(AbstractBaseUser, PermissionsMixin):
         max_length=1,
     )
 
-    utilities = models.ManyToManyField(
-        Utility,
+    utilities = models.ManyToManyField(Utility,
         blank=True,
         related_name="users",
     )
diff --git a/src/django/api/models/utility.py b/src/django/api/models/utility.py
index b3b0f73..5b62ecb 100644
--- a/src/django/api/models/utility.py
+++ b/src/django/api/models/utility.py
@@ -4,7 +4,7 @@ from django.db import models
 
 from .state import State
 
-RALEIGH = Point(-78.6382, 35.7796)
+RALEIGH = Point(    -78.6382, 35.7796)
 
 
 __all__ = ["Utility"]
```

And I run `git commit -m TEST`, it only fixes one file at a time:

```
Checking api/models/boundary.py
would reformat api/models/boundary.py

Oh no! 💥 💔 💥
1 file would be reformatted.
Formatting with black: api/models/boundary.py
reformatted api/models/boundary.py

All done! ✨ 🍰 ✨
1 file reformatted.
```

Also, despite adding the `--loglevel warn` flag for prettier, which works correctly when I run `./scripts/format`, I still see the noisy prettier output in the git hook. Would be good to silence that on noop.

jacobtylerwalls:
Ah, thanks, I tested that case earlier in development and then neglected to test it again on latest changes. Turned out Docker was taking over `stdin`, so I needed to find another way to split strings in bash (i.e. not using `while`).

1e0cd6c can now format multiple python files and asks for quieter output of prettier and black.

> Also, despite adding the --loglevel warn flag for prettier, which works correctly when I run ./scripts/format, I still see the noisy prettier output in the git hook. Would be good to silence that on noop.

I see this output now:
```sh
iow-boundary-tool % git commit -am 'test'
format...................................................................Failed
- hook id: format
- files were modified by this hook

yarn run v1.22.19
$ /usr/local/src/node_modules/.bin/prettier --loglevel warn --write src/
Done in 3.02s.
```

LMK if you see something more verbose than that when committing!

rajadain:
Will take another look

rajadain (commented):
This is looking good!

One final thing I noticed is that it may be faster to give the format and check commands the list of all files, instead of doing it line by line. That would mean that the Docker container would only be need to be brought up once, rather than for each file:

```diff
diff --git a/scripts/format b/scripts/format
index 7d255c1..4f3145d 100755
--- a/scripts/format
+++ b/scripts/format
@@ -24,31 +24,28 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         # Format Django app
 
         # Get a list of changed ".py" files and trim "src/django"
-        changed_files=$(git status --porcelain | grep ".py" | awk -F'src/django/' '{print $2}')
+        changed_files=$(git status --porcelain | grep ".py" | awk -F'src/django/' '{print $2}' | xargs)
         if [[ -n "${changed_files}" ]]; then
-            export IFS=$'\n'
-            for line in ${changed_files}; do
-                # Run isort --check-only before fixing to ensure a failure code is emitted
-                if ! docker-compose \
-                    run --rm --no-TTY --no-deps --entrypoint isort django "${line}" --check-only; then
-
-                    docker-compose \
-                        run --rm --no-TTY --no-deps --entrypoint isort django "${line}"
-                fi
-                # Run black with --check before fixing to ensure a failure code is emitted
-                if ! docker-compose \
+            # Run isort --check-only before fixing to ensure a failure code is emitted
+            if ! docker-compose \
+                run --rm --no-TTY --no-deps --entrypoint isort django ${changed_files} --check-only; then
+
+                docker-compose \
+                    run --rm --no-TTY --no-deps --entrypoint isort django ${changed_files}
+            fi
+            # Run black with --check before fixing to ensure a failure code is emitted
+            if ! docker-compose \
+                run --rm --no-TTY --no-deps --entrypoint black django \
+                --skip-string-normalization \
+                --extend-exclude migrations ${changed_files} \
+                --quiet --check; then
+
+                docker-compose \
                     run --rm --no-TTY --no-deps --entrypoint black django \
                     --skip-string-normalization \
-                    --extend-exclude migrations "${line}" \
-                    --quiet --check; then
-
-                    docker-compose \
-                        run --rm --no-TTY --no-deps --entrypoint black django \
-                        --skip-string-normalization \
-                        --extend-exclude migrations \
-                        --quiet "${line}"
-                fi
-            done
+                    --extend-exclude migrations \
+                    --quiet ${changed_files}
+            fi
         fi
     fi
 fi
```

jacobtylerwalls:
Thanks, your diff works great. I thought I determined that black didn't take a list of files, but I should have been more skeptical of such a missing feature. I must have mistook a failure to specify the correct path in the docker container for black not accepting a list of files.

scripts/format
```diff
@@ -0,0 +1,47 @@
...
+        usage
+    else
+        # Format Django app
+
+        # Get a list of changed ".py" files and trim "src/django"
+        changed_files=$(git status --porcelain | grep ".py" | awk -F'src/django/' '{print $2}' | xargs)
+        if [[ -n "${changed_files}" ]]; then
+            # Run isort --check-only before fixing to ensure a failure code is emitted
+            if ! docker-compose \
+                run --rm --no-TTY --no-deps --entrypoint isort django "${changed_files}" --check-only; then
```
rajadain, line 32: Since this is now a list of multiple files, I don't think this should be quoted anymore. Getting this error:

```
isort.exceptions.InvalidSettingsPath: isort was told to use the settings_path: /usr/local/src/api/models/boundary.py api/serializers/boundary.py api/views as the base directory or file that represents the starting point of config file discovery, but it does not exist.
```

Removing the quotes everywhere fixes the run.

rajadain (changes requested):

scripts/format
jacobtylerwalls: Ah, then I will have to silence the `shellcheck` warning, which was failing the CI run. I'll look into if they have pragmas for that.

rajadain:
Taking a final look

rajadain (approved):
+1 tested, this looks good! One optional recommendation I'll make is regarding the `format` script:

https://github.com/azavea/iow-boundary-tool/blob/68b19e0bd8cefd5fb50ad1630f8a209b555d34af/scripts/format#L32-L53

Currently, there are four `docker compose run --rm django` calls, which means that a new container is spun up, does one thing, then killed, four times.

Given that Docker is a heavier lift on macOS than Linux, doing so less often would be preferred.

I would recommend trying to see if we can spin up the `django` container once, and run a script _within_ it to do all this work. It will be faster and more efficient, with fewer Docker initializations. However, the django format script will have to be moved into the django folder and container, which may be a little awkward feeling as most scripts therein are Python, not Bash.

A timeboxed evaluation of such a technique is recommended.

Great work on this!! 👏👏

jacobtylerwalls:
Thanks for the terrific suggestions!

PR #210, opened 2022-11-15
Memoize shape update functions

### LINKED ISSUE #207: Draw Page: Autosave Interrupts Drawing Mid-drag
## Overview

When editing a shape, if the shape is autosaved while the user is dragging a corner, the shape resets and their drag is lost. This can frustrating when doing detailed work, where one may often drag a corner for ~3s or more:

https://user-images.githubusercontent.com/1430060/201749555-c3759787-46ba-499b-a4de-81cfeb1d26b8.mp4

Find a way around this so that the user is not interrupted by autosaves.

mstone121:
## Overview

Memoizing these functions allows the `useTrailingDebounceCallback` to memoize its returned callback. This prevents the editing polygon layer from re-rendering each time its containing hook is called.

Closes #207 

### Notes

The reference images (which also use `useTrailingDebounceCallback`) don't need this fix because they aren't reactive to state updates.

## Testing Instructions

- Go the draw page in draft mode
- Edit the shape
- Before the debounced save shape request happens, click and hold one of the verticies
  - [x] Ensure the vertex doesn't jump back when the shave shape request happens

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix view submission details link destination [#174](https://github.com/azavea/iow-boundary-tool/pull/174)
 - Fix submit boundary [#195](https://github.com/azavea/iow-boundary-tool/pull/195)
 - Fix contributor welcome redirect [#197](https://github.com/azavea/iow-boundary-tool/pull/197)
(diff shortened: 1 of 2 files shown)


rajadain (approved):
+1 tested, nice fix!! I had tried to wrap `useTrailingDebounceCallback` inside a `useCallback` a while ago when attempting something else, and it wouldn't let me put a hook inside `useCallback`. This is the right way to do it. Great work!

PR #218, opened 2022-11-23
Add Support for Exporting Boundaries

### LINKED ISSUE #93: Administrator Role Support in front-end
## Overview

Ensure that an Administrator can act as Validator in all cases. Ensure they can also act as Contributor in some cases (ability to edit submissions, etc).

### LINKED ISSUE #172: NavBar Should Show Submission Utility, Not Selected Utility
## Overview

The NavBar should always show the submission's utility, not the one that the user has selected:

https://user-images.githubusercontent.com/1430060/199102062-3c8b8710-865d-4a90-8ba2-c36adc984b24.mp4

### LINKED ISSUE #184: Automatically determine utility at utillity specific URLs
## Overview

Going straight to a utility specific URL (e.g. `/submissions/4` or `/draw/4`) still shows the utility selector. This can be annoying and potentially confusing. If a user selects a utility not related to that of the URL's item, then the user will still see the URL item, but will have a different utility selected.

### Describe the solution you'd like

Automatically set the utility when the user visits one of these URLs

We should wrap all of the utility specific routes in a component which defers rendering of its children until the utility is set. This would probably be easier to do if we moved the draw page "into" submissions: `/submissions/4/draw`.

### LINKED ISSUE #211: Hide Add Map Button for Validators
## Overview

If the logged in user is a Validator, the Add Map button should be hidden:

image

### LINKED ISSUE #216: Add Ability to Export Boundaries
## Overview

Add a button in the UI for administrators that hits an endpoint only they can access, that exports all the approved boundaries with their shapes as a geojson.

Since the amount of data can potentially be very large, design a query that generates the geojson in the database and directly send / stream that data to the user, rather than constructing it in Django.

rajadain:
## Overview

Adds two new endpoints to export boundaries: one for only approved boundaries, and one for all boundaries. Adds a button in the front-end for administrators to do so.

Also cleans up some minor issues.

Closes #216 
Closes #172 
Closes #93 
Closes #211 
Closes #184 

### Demo

image

image

### Notes

5e6fbf2d8f6c7318bceff46209fbf3d1f3c34289 was done when I thought that the button would live in the List NavBar, but in the end it made more sense to place it on the main page itself. It can be excluded, but I think it is still valuable, so kept it in.

## Testing Instructions

- Checkout this branch and `server`
- Login as an Administrator
- Go to the Submissions List
  - [x] Ensure you see a button to export boundaries
  - [x] Ensure you are able to export all boundaries
  - [x] Ensure you are able to export approved boundaries
- Log out, and login as a Contributor
- Open this URL: http://localhost:4545/submissions/1
- ~~Select "Other Utility" in the selector dropdown~~
  - [ ] ~~Ensure the loaded page says "Azavea Test Utility" in the top-left, not "Other Utility"~~
  - [x] Ensure you do not see the utility selector dropdown
  - [x] Ensure the loaded page says "Azavea Test Utility" in the top-left
- Open this URL: http://localhost:4545/submissions
  - [x] Ensure you do see the utility selector dropdown
  - [x] Ensure you are taken to the appropriate submissions list after

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/components/UtilityGuard.js
+++ b/src/app/src/components/UtilityGuard.js
@@ -1,11 +1,13 @@
 import { useEffect, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
-import { useNavigate } from 'react-router-dom';
+import { useLocation, useNavigate } from 'react-router-dom';
 import { Button, Text } from '@chakra-ui/react';
 
+import { useGetBoundaryDetailsQuery } from '../api/boundaries';
 import apiClient from '../api/client';
 import LoginForm from '../components/LoginForm';
 import { UtilitySelector } from './UtilityControl';
+import CenteredSpinner from './CenteredSpinner';
 
 import { logout, setUtilityByPwsid } from '../store/authSlice';
 
@@ -25,6 +27,15 @@ export default function UtilityGuard({ children }) {
 
 function ContributorUtilityGuard({ children }) {
     const dispatch = useDispatch();
+    const { pathname } = useLocation();
+
+    const boundaryId = pathname.match(/^\/submissions\/(\d+)/)?.[1];
+
+    const {
+        isLoading,
+        data: boundary,
+        error,
+    } = useGetBoundaryDetailsQuery(boundaryId, { skip: !boundaryId });
 
     const {
         user: { utilities },
@@ -39,11 +50,28 @@ function ContributorUtilityGuard({ children }) {
         utilities.length === 1 ? utilities[0].pwsid : null;
 
     // Set the utility automatically if the user has only one
+    // Or set it to the given boundaryId's
     useEffect(() => {
         if (!utility && utilityToAutoSelect) {
             dispatch(setUtilityByPwsid(utilityToAutoSelect));
         }
-    }, [utility, utilityToAutoSelect, dispatch]);
+
+        if (boundaryId && !isLoading && !error && boundary) {
+            dispatch(setUtilityByPwsid(boundary?.utility?.pwsid));
+        }
+    }, [
+        boundaryId,
+        isLoading,
+        boundary,
+        error,
+        utility,
+        utilityToAutoSelect,
+        dispatch,
+    ]);
+
+    if (isLoading) {
+        return <CenteredSpinner />;
+    }
 
     if (utility) {
         return children;
--- a/src/django/api/views/export.py
+++ b/src/django/api/views/export.py
@@ -1,0 +1,64 @@
+from datetime import datetime
+
+from django.db import connection
+from django.http import HttpResponse
+from rest_framework.decorators import api_view, permission_classes
+from rest_framework.permissions import IsAuthenticated
+
+from api.permissions import UserIsAdmin
+
+
+@api_view(["GET"])
+@permission_classes([IsAuthenticated, UserIsAdmin])
+def boundaries_approved(request):
+    query = """
+    WITH features AS (
+        SELECT json_build_object(
+            'type',       'Feature',
+            'id',         b.id,
+            'geometry',   ST_AsGeoJSON(shape)::json,
+            'properties', json_build_object(
+                'name',            u.name,
+                'pwsid',           u.pwsid,
+                'city',            u.address_city,
+                'state',           u.state_id,
+                'drafted_at',      s.created_at,
+                'drafted_by',      du.email,
+                'submitted_at',    s.submitted_at,
+                'submitted_by',    su.email,
+                'submitted_notes', s.notes,
+                'approved_at',     a.approved_at,
+                'approved_by',     au.email,
+                'status',          'Approved'
+            )
+        ) AS feature
+        FROM api_approval a
+            INNER JOIN api_user au ON a.approved_by_id = au.id
+            INNER JOIN api_submission s ON a.submission_id = s.id
+            INNER JOIN api_user du ON s.created_by_id = su.id
+            INNER JOIN api_user su ON s.submitted_by_id = su.id
+            INNER JOIN api_boundary b ON s.boundary_id = b.id
+            INNER JOIN api_utility u ON b.utility_id = u.id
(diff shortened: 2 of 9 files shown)


src/django/api/views/export.py
```diff
@@ -0,0 +1,144 @@
...
+                FROM api_review
+                GROUP BY submission_id) lr ON s.id = lr.submission_id
+            LEFT JOIN api_review r ON r.id = lr.id
+            LEFT JOIN api_user ru ON r.reviewed_by_id = ru.id
+            LEFT JOIN (
+                SELECT MAX(id) AS id, submission_id
+                FROM api_approval
+                GROUP BY submission_id) la ON s.id = la.submission_id
+            LEFT JOIN api_approval a ON a.id = la.id
+            LEFT JOIN api_user au ON a.approved_by_id = au.id
```
mstone121, line 122: `LATERAL` subqueries can reference columns from earlier in the `FROM` clause. This lets us use the date to find the latest submission/review rather than the id where the order might not be guaranteed.

More info can be found in `7.2.1.5. LATERAL Subqueries` of [7.2.1. The FROM Clause](https://www.postgresql.org/docs/12/queries-table-expressions.html#QUERIES-FROM).
 
```suggestion
            INNER JOIN api_utility u ON u.id = b.utility_id
            LEFT JOIN LATERAL (
                SELECT * FROM api_submission
                WHERE boundary_id = b.id
                ORDER BY created_at DESC
                LIMIT 1
            ) s ON TRUE
            LEFT JOIN api_user du ON s.created_by_id = du.id
            LEFT JOIN api_user su ON s.submitted_by_id = su.id
            LEFT JOIN LATERAL (
                SELECT * FROM api_review
                WHERE submission_id = s.id
                ORDER BY created_at DESC
                LIMIT 1
            ) r ON TRUE
            LEFT JOIN api_user ru ON r.reviewed_by_id = ru.id
            LEFT JOIN LATERAL (
                SELECT * FROM api_approval
                WHERE submission_id = s.id
                ORDER BY approved_at DESC
                LIMIT 1
            ) a ON TRUE
            LEFT JOIN api_user au ON a.approved_by_id = au.id
```

mstone121 (approved):
Looks good! Everything is working as expected.

It's unfortunate that Django does not have more support for database views (perhaps due to vendor differences) or else we wouldn't have to replicate status/latest logic.

src/django/api/views/export.py
```diff
@@ -0,0 +1,144 @@
...
+            'type',       'Feature',
+            'id',         b.id,
+            'geometry',   ST_AsGeoJSON(shape)::json,
+            'properties', json_build_object(
+                'name',            u.name,
+                'pwsid',           u.pwsid,
+                'city',            u.address_city,
+                'state',           u.state_id,
+                'drafted_at',      s.created_at,
+                'drafted_by',      du.email,
```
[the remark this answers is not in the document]
rajadain: Good point. I made the all boundaries endpoint afterwards, will include them in the approved one too, and generally align the data structures.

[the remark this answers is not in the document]
rajadain: [4a6815c](https://github.com/azavea/iow-boundary-tool/pull/218/commits/4a6815cb98a0a61c66afcf62efa7b176dc262829)

src/django/api/views/export.py
rajadain: Ah interesting! I hadn't heard of this before. Still don't quite have an intuition after reading up on it. I think of subqueries are row-level operations, and joins as table-level operations, so it's strange to have a subquery in a join statement. I timed them in psql though and the LATERAL JOIN appears to be running faster (???) which does not intuit well.

Nevertheless, the results are solid, and the reasoning makes sense, so I'm going with it.

[ed267d5](https://github.com/azavea/iow-boundary-tool/pull/218/commits/ed267d581e48ea8ff465e25a6fba7f5096d4dca8)

rajadain:
Thanks for taking a look and your suggestions! I was able to solve #184 after all, updated the testing instructions to reflect that.

I also reverted the commit fixing #172 since that is now fixed by this new work. Will end up dropping that before merging.

rajadain:
Would appreciate another quick look for the later additions to this PR

src/app/src/components/UtilityGuard.js
```diff
@@ -25,6 +27,25 @@ export default function UtilityGuard({ children }) {
...
+
+    if (pathname.startsWith('/submissions/')) {
+        boundaryId = pathname.substring('/submissions/'.length);
+        if (pathname.endsWith('/draw')) {
+            boundaryId = boundaryId.substring(
+                0,
+                boundaryId.length - '/draw'.length
+            );
+        }
+    }
```
mstone121, line 42:
```suggestion
    const boundaryId = pathname.match(/^\/submissions\/(\d+)/)?.[1];
```

mstone121 (approved):
👍 Thanks for adding the utility auto-select!

src/app/src/components/UtilityGuard.js
rajadain: Perfect. Fixed in [b555cbc](https://github.com/azavea/iow-boundary-tool/pull/218/commits/b555cbc95cb4f6234bde07de95fc5e8bf577a613)

rajadain:
Rebased on latest `develop`, will merge once green

src/app/src/components/UtilityGuard.js
```diff
@@ -25,6 +27,15 @@ export default function UtilityGuard({ children }) {
 
 function ContributorUtilityGuard({ children }) {
     const dispatch = useDispatch();
+    const { pathname } = useLocation();
+
+    const boundaryId = pathname.match(/^\/submissions\/(\d+)/)?.[1];
```
rajadain, line 32: We use this here instead of `useBoundaryId` because this component is above the `Route` that declares that parameter

rajadain:
Thanks for the multiple thorough reviews and all the great suggestions!

PR #234, opened 2023-05-19
Add Ingested Tile Support, National Municipal Layer

### LINKED ISSUE #233: Add National Municipal Boundary Layer
## Overview

Kyle has provided us with a National Municipal Boundary layer: https://drive.google.com/file/d/1ohjue3xNeypf1C8vU2qRCA7raGTAfzkM/view?usp=share_link. This is a 56 MB GeoJSON file.

Replace the current NC-specific Municipal Boundary Layer with this one.

### Is your feature request related to a problem? Please describe.

Since the application is nation scale, having a states-specific boundary layer is confusing. We should use the national one.

### Describe the solution you'd like

The state-specific municipal boundary layer is replaced with the national one. The styling and performance is similar to before.

### Additional Context

Given that this file is much bigger than the other one, it is likely that just serving it directly will not work.

Cut this into Vector Tiles using https://github.com/mapbox/tippecanoe and [the recommended options for sparse data](https://github.com/mapbox/tippecanoe#continuous-polygon-features-states-and-provinces-visible-at-all-zoom-levels). Use `-e` to generate Protobuf tiles in ZXY format. Script this import.

Upload those tiles to a new S3 bucket, that should be added to terraform.

Add a [vector tiles plugin](https://leafletjs.com/plugins.html#vector-tiles) to Leaflet.

Load the tiles and style them to be similar to the current layer.

rajadain:
## Overview

We need to display vector datasets that are too large to be loaded into the browser, such as the National Municipal Layer which is a ~56MB GeoJSON file. To render them in a performant manner, we convert the layer into Vector Tiles and host it on S3 + CloudFront, and load it from there using the [Leaflet VectorGrid plugin](https://leaflet.github.io/Leaflet.VectorGrid/vectorgrid-api-docs.html).

This PR adds the scripts to generate the vector tiles, the infrastructure to host it, the wiring to configure it, and the front-end components to render it. It also adds documentation around how to ingest future datasets.

Closes #233 

### Demo

https://github.com/azavea/iow-boundary-tool/assets/1430060/93748a57-1d9c-431f-a7d8-60a690288639

### Notes

~~Since we're using a CloudFront distribution to serve the tiles, it is likely we'll need to add an invalidate mechanism for whenever we publish new tiles to ensure that the latest tiles are always served.~~ Added in 337285f.

The previous Municipal Boundary layer had a name for every boundary that was shown as labels at certain zoom levels. This dataset does not have those labels, so none are shown.

## Testing Instructions

- Check out this branch and `update` and `server`
- Go to http://localhost:4545/ and login as `a1@azavea.com`
- Open a submission, go to its Map view
- Play around with the Municipal Boundary layer
  - [x] Ensure toggling it on and off works
  - [x] Ensure it does not interfere with the Parcel layer and is generally shown in the right spot (above the parcels, beneath the user drawn boundary)
  - [x] Ensure you can see municipal boundaries from around the country

## Checklist

- [x] CI passes after rebase


data/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+The tiles are stored in S3 and served via CloudFront Distributions, which have friendly URLs:
+
+| Environment | S3 Bucket                        | URL                                             |
+|-------------|----------------------------------|-------------------------------------------------|
+| Staging     | `iow-staging-tiles-us-east-1`    | https://tiles.staging.iow.azavea.com/           |
+| Production  | `iow-production-tiles-us-east-1` | https://tiles.boundarysync.internetofwater.app/ |
+
+### Ingest Script
+
```
rachelekm, line 17:
```suggestion
Confirm `ogr2ogr` and `tippecanoe` installed locally:
```bash
ogr2ogr -v
tippecanoe -v
```

I think this is obvious enough from the below scripts, but could be helpful as a first step for the sake of explicit instruction

src/app/src/components/Layers/MunicipalBoundariesLayer.js
```diff
@@ -1,68 +1,37 @@
...
-                                  basemapType === 'satellite'
-                                      ? ' muni-label-light'
-                                      : ''
-                              }`,
-                              pane: PANES.MUNICIPAL_BOUNDARY_LABELS.label,
-                          });
-                      }
-                    : null,
-                interactive: false,
+function VectorGrid({ url, maxNativeZoom = 14 }) {
```
rachelekm, line 15: Elsewhere the `maxNativeZoom` is 12 and that's the max zoom defined when running `tippecanoe` command, is 14 the default because that's more standard for tiles?

src/django/web/templates/web/environment.js
```diff
@@ -0,0 +1 @@
+window.ENVIRONMENT = Object.freeze({{ environment | safe }});
```
rachelekm, line 1: For my own learning can you walk me through this line? I think I understand this as making the `environment` property immutable which we use as an environment var to construct the right tiles url, but I'm not sure where `safe` comes from.

rachelekm (approved):
Really great job, there was a lot to this PR but it felt easy to gain context on/figure out how parts related to each other thanks to your README contribution and helpful notes! I tried out different parts of the country at varying zoom levels/toggling on and off the municipal boundary and everything functioned as expected. I left a couple of clarifying comments (mostly for my own learning!) but I think this looks great 👍

data/README.md
rajadain: Addressed in f751691

src/app/src/components/Layers/MunicipalBoundariesLayer.js
rajadain: Good catch! I saw `maxNativeZoom: 14` in the example here: https://leaflet.github.io/Leaflet.VectorGrid/vectorgrid-api-docs.html#vectorgrid-protobuf and mentally internalized it as the default value, which it isn't: it's just a demo value.

Going to switch to `12` here for consistency.

The value of `12` was chosen after some trial and error to keep the tileset size low while also having enough detail for when users are drawing their shapes. After `12` I saw diminishing returns on increased size. At `12` the tileset is about less than 400MB, and costs about $0.05 / month to host with 10K requests, which seemed acceptably affordable.

I'll add that bit to the Ingest README.

rajadain: Fixed in a254405 and ce02f5e

src/django/web/templates/web/environment.js
rajadain: This file is really a Jinja template, used by Django to render a page. Usually we use Jinja syntax in HTML files, and it's more obvious. Here we're using it in a JavaScript file.

The template is populated with values here:

https://github.com/azavea/iow-boundary-tool/blob/ce02f5e9a5930ce7a3ed4b6920f1712946e3d87d/src/django/web/views.py#L17-L24

The `environment` in the template is populated using the `context` that is provided to the `render` function.

The `safe` is a Jinja filter, used to escape HTML so that the output is not interpreted and executed by the browser, but is taken literally. This changes `<` into `&lt;`, `&` into `&amp;`, etc. For more detail, see here: https://jinja.palletsprojects.com/en/3.1.x/templates/#html-escaping

rajadain:
Thanks for reviewing!

src/django/web/templates/web/environment.js
rachelekm: Oh! Thank you for explaining that makes sense!

